PR #19, opened 2018-03-09
Support locked problems

TsFreddie:
Added support for locked problems.
The regex is changed to match favorite, locked and state:
```
★ 🔒 ✔ [760] Find Anagram Mappings                                        Easy   (75.23 %)
```
The lock icon is grabbed from EmojiOne v2.0, please change it or add a license.

Please review the regex before merge.


jdneo:
Hi @TsFreddie Thank you for the contribution! Looks nice!

Just interested about the intension of supporting locked problem. Could you please explain more about this feature to me?

TsFreddie:
I should say paid problems.

leetcode-cli does not give you all the problems with `leetcode list -q L` even if you are a premium user of leetcode.

image

It would be nice to be able to submit those problems as well.

jdneo:
@TsFreddie Oh, got your idea. That means: although the icon is a lock, but the paid user still can open the problem and try to solve it. Am I right?

I don't have a premium account so I was not aware of that. 😄

TsFreddie:
Exactly. leetcode-cli isn't very descriptive about what it means by "locked".

jdneo:
Good to know that. Thanks for pointing out. 👍 

I still have some stuff need to do. Will review the PR in detail ASAP!

src/commands/list.ts
```diff
@@ -7,6 +7,8 @@ import { executeCommand } from "../utils/cpUtils";
 import { DialogType, promptForOpenOutputChannel } from "../utils/uiUtils";
 
 export interface IProblem {
+    favorate: boolean;
```
jdneo: typo: favorite

src/commands/list.ts
```diff
@@ -19,19 +21,23 @@ export async function listProblems(channel: vscode.OutputChannel): Promise<IProb
         if (leetCodeManager.getStatus() === UserStatus.SignedOut) {
             return [];
         }
-        const result: string = await executeCommand(channel, "node", [leetCodeBinaryPath, "list", "-q", "L"]);
+        const leetCodeConfig: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("leetcode");
+        const showLocked = leetCodeConfig.get<string>("showLocked");
```
jdneo: nit: add a type definition: `const showLocked: boolean = ...`

package.json
```diff
@@ -129,6 +129,11 @@
             {
                 "title": "LeetCode",
                 "properties": {
+                    "leetcode.showLocked": {
+                        "type": "boolean",
+                        "default": true,
```
jdneo: Suggest we set the default value to false, since I think there are more common users than the premium ones.

jdneo:
@TsFreddie Just some small changes, thanks.

jdneo:
Great! Thanks.

ismailhkose:
image

PR #57, opened 2018-08-13
support leetcode cn

### LINKED ISSUE #50: Support leetcode-cn
[LeetCode 中国](https://leetcode-cn.com/) is the Chinese version of LeetCode. It provides a more comfortable experience for Chinese users, I think.

However, data of the universal version and the Chinese version of LeetCode is not shared. If we want, we have to sync them manually. So can we choose which version to use? 😃 

Thanks!

BTW, I have learned that this extension is based on the [leetcode-cli](https://github.com/skygragon/leetcode-cli) project. Maybe I had better open an issue there...

jdneo:
resolve #50


package.json
```diff
@@ -216,7 +238,7 @@
         "vscode:prepublish": "npm run compile",
         "compile": "tsc -p ./",
         "watch": "tsc -watch -p ./",
-        "postinstall": "node ./node_modules/vscode/bin/install",
+        "postinstall": "node ./node_modules/vscode/bin/install && node ./node_modules/leetcode-cli/bin/leetcode plugin -d leetcode.cn",
```
jdneo: should be `-i`

Issue #67, opened 2018-10-07
support lable search ，for example array，map ，greedy algorithms


luoye2017:
rt

Vigilans:
Would be better if we can list the problems by tag in the problem view.
The current `Easy-Medium-Hard` convention seems not fully making use of the problem view space.

PR #94, opened 2019-01-23
Enhanced Tree view

Vigilans:
## Introduction
This PR restructures the left side panel's tree view to support following functions:
- [x] View problems by difficulty, tag or company #67 
- [x] View favorite problems
- [x] Option to hide accepted questions (won't be hidden in `Favorite` view)

## Details
* Company and tag filtering is achieved through the `leetcode-cli` [company](https://github.com/skygragon/leetcode-cli-plugins/blob/master/plugins/company.js) plugin.
* Problems with no tag or company knowledge are labeled as `Unknown`.
* Solved problems can be hidden through `leetcode.hideSolved` setting, which defaults to false.

## Demonstration
Difficulty, tag, company or favorite:
image

Problems by tag:
image

Problems by company:
image

Favorite problems:
image


jdneo:
@Vigilans Thank you for contribution. I will look it later.

Also noticed that there are some lint errors according to the CI output, would you mind to fix them first?

Vigilans:
The lint errors have been fixed.

jdneo:
Hi @Vigilans,

 Is there any reason to add accepted/not accepted categories into the explorer?

Vigilans:
The Not-AC category shows problems which are submitted but not accepted.  
For AC category, I wonder if there will be some mechanisms like retrieving all accepted codes of one problem. For now, this category seems redundant...

If you think them not necessary, I will try to revert this commit.

jdneo:
Hmm, I think it's a little bit redundant somehow.

I better approach might be: having an extension setting that specifies if the explorer will only show unresolved problems or all the problems, then this setting will apply to all the categories.

To retrieving all accepted codes, this can be another action in the explorer context menu, which is another story.

taoweicn:
@jdneo Yeah, the feature I mentioned. 😄 #95

Vigilans:
I see.
---
There are some other changes I want to contribute, maybe I should open another issue to query you for your opinions_(눈_ 눈」∠)_

Vigilans:
@taoweicn I will work on this feature later.

Vigilans:
@jdneo The `hiding solved` feature is implemented with only 15 lines of code in 2 files, and is part of the tree view.
Personally I think there is not much burden in reviewing this feature, unless more requirements are introduced. Next time I will take care of it when committing new changes_(:зゝ∠)_

src/commands/list.ts
```diff
@@ -7,15 +7,20 @@ import { leetCodeManager } from "../leetCodeManager";
...
+    state: ProblemState.Unknown,
+    id: "",
+    name: "",
+    difficulty: "",
+    passRate: "",
+    companies: [] as string[],
+    tags: [] as string[],
+};
+
+export type IProblem = typeof IProblemDefault;
```
jdneo: It would be better to put the `IProblem` definition and `IProblemDefault` to other places instead of in this file.
Let's put them into `shared.ts`, I'll refactor this part later.

src/leetCodeExecutor.ts
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 import * as cp from "child_process";
+import * as fs from "fs";
```
jdneo: Use `fs-extra`

src/leetCodeExecutor.ts
```diff
@@ -100,6 +110,19 @@ class LeetCodeExecutor {
         }
     }
 
+    public async getCompaniesAndTags(): Promise<{ companies: { [key: string]: string[] }, tags: { [key: string]: string[] } }> {
+        // preprocess the plugin source
+        const componiesTagsPath: string = path.join(await leetCodeExecutor.getLeetCodeRootPath(), "lib", "plugins", "company.js");
+        const componiesTagsSrc: string = (await fs.readFileSync(componiesTagsPath, "utf8")).replace(
```
jdneo: We can use `readFile` from `fs-extra`.

src/leetCodeExecutor.ts
```diff
@@ -100,6 +110,19 @@ class LeetCodeExecutor {
...
+            "module.exports = plugin",
+            "module.exports = { COMPONIES, TAGS }",
+        );
+        // require plugin from modified string
+        const requireFromString: (src: string, path: string) => any = require("require-from-string");
```
jdneo: Can we import the module?

For example, `npm i --save-dev @types/require-from-string`
Then import it instead of using require

src/leetCodeExplorer.ts
```diff
@@ -7,6 +7,8 @@ import * as list from "./commands/list";
 import { leetCodeManager } from "./leetCodeManager";
 import { ProblemState } from "./shared";
 
+export type Category = "Difficulty" | "Tag" | "Company";
```
jdneo: Can we use enum instead?

src/leetCodeExplorer.ts
```diff
@@ -64,57 +73,100 @@ export class LeetCodeTreeDataProvider implements vscode.TreeDataProvider<LeetCod
...
-                problems.push(problem);
-            } else {
-                this.treeData.set(problem.difficulty, [problem]);
+        this.treeData = {
+            Difficulty: new Map(),
+            Tag: new Map(),
+            Company: new Map(),
+            Favorite: [],
+        };
+        for (const problem of await list.listProblems()) {
```
jdneo: I'm thinking that we can filter the accepted problems in the method `list.listProblems()`. If `hideSolved === true` then only return unaccepted problems.

By doing this, we decouple the `this.leetCodeConfig` with the explorer

src/leetCodeExplorer.ts
```diff
@@ -64,57 +73,100 @@ export class LeetCodeTreeDataProvider implements vscode.TreeDataProvider<LeetCod
...
+            // Add favorite problem, no matter whether it is solved.
+            if (problem.favorite) {
+                this.treeData.Favorite.push(problem);
+            }
+            // Hide solved problem in category view is specified in option.
+            if (problem.state === ProblemState.AC && this.leetCodeConfig.get<boolean>("hideSolved")) {
+                continue;
+            }
+            // Add problems according to category
+            const categories: Array<[Category, string[]]> = [
```
jdneo: Can we use Map here?

use index to represent the `parent-children` relationship is confusing

src/leetCodeExplorer.ts
```diff
@@ -64,57 +73,100 @@ export class LeetCodeTreeDataProvider implements vscode.TreeDataProvider<LeetCod
...
+                ["Difficulty", [problem.difficulty]],
+                ["Tag", problem.tags],
+                ["Company", problem.companies],
+            ];
+            for (const [parent, children] of categories) {
+                for (let subCategory of children) {
+                    // map 'first-second' to 'First Second'
+                    subCategory = subCategory.split("-").map((c: string) => c[0].toUpperCase() + c.slice(1)).join(" ");
```
jdneo: `first-second` is not that bad.

Cuz current logic could handle `a-b` but not `a-b-c`

Just leave `first-second` as it is for now.

src/leetCodeExplorer.ts
```diff
@@ -64,57 +73,100 @@ export class LeetCodeTreeDataProvider implements vscode.TreeDataProvider<LeetCod
         }
 
         const idPrefix: number = Date.now();
-        return {
+        return { // element.id -> parent node name, element.name -> node name
```
jdneo: Why we save parent node name into `element.id`?

src/leetCodeExplorer.ts
Vigilans: What does `a-b-c` refer to? The code here could transform `breadth-first-search` to `Breadth First Search`.

src/leetCodeExplorer.ts
Vigilans: We could use Map here this way:
```typescript
const categories: Map<Category, string[]> = new Map<Category, string[]>([
    ["Difficulty", [problem.difficulty]],
    ["Tag", problem.tags],
    ["Company", problem.companies],
]);
```
Btw, a better way would be to use `Object.entries` here, but it is not supported by es6 in the `tsconfig.json`.

src/leetCodeExplorer.ts
Vigilans: There should be some field to store the parent node name when it is not problem node...since in previous commits you simply set `node.id` and `node.name` the same value, I decided to reuse the `id` field to represent some other value.

The reasons for the necessity of parent node name are:
1. We should be able to retrieve data from the `LeetCodeTreeDataProvider.treeData`.
2. One problem will now be added as `LeetCodeNode` at least 3 times(in `Difficulty`, `Tag` and `Company`), we should be able to distinguish them, hence the following naming rule:
```typescript
id: `${idPrefix}.${element.id}.${element.name}`,
```
(`idPrefix = Date.now()` just returns same value throughout initialization process)

src/leetCodeExecutor.ts
Vigilans: Yeah, the `definitelyTyped` indeed provides type for `require-from-string`.
`import * as requireFromString from "require-from-string";` successfully worked.

src/leetCodeExplorer.ts
Vigilans: In my opinion, `Favorite` group should show all starred problems no matter whether it's accepted or not. Filtering problems in `list.listProblems()` seems too early for this requirement.

src/leetCodeExecutor.ts
Vigilans: Something went wrong with my usage. `fs-extra` is correct.

src/leetCodeExplorer.ts
jdneo: Hmm, ok, just keep it there first.

src/leetCodeExplorer.ts
jdneo: Let's use Map here.

src/leetCodeExplorer.ts
jdneo: My bad, it's fine.

src/leetCodeExplorer.ts
jdneo: Maybe we could add a filed in the `LeetCodeNode` called 'parentName' to store the parent node info.

src/leetCodeExplorer.ts
jdneo: Have you tried this:

```ts
export enum Category {
    Difficulty = "Difficulty",
    Tag = "Tag",
    Company = "Company",
}
```

jdneo: No need to call `.toString()` here.

Whenever you want to use the Category value as a string, just using `Category.xxxx`.

By doing this, you will get the value as a string meanwhile keep the typedef.

I tried using enum here and everything works fine.

Vigilans: Let's make the `Category` enum type:
```typescript
export enum Category {
    Difficulty = "Difficulty",
    Tag = "Tag",
    Company = "Company",
}
```

Then rewrite [line 147](https://github.com/jdneo/vscode-leetcode/blob/d9abd5f68a92bc7679066ea35bd342fd80a599ea/src/leetCodeExplorer.ts#L147) this way:
```typescript
const categories: Map<Category, string[]> = new Map<Category, string[]>([
    [Category.Difficulty, [problem.difficulty]],
    [Category.Tag, problem.tags],
    [Category.Company, problem.companies],
]);
```

Follows the iteration:
```typescript
for (const [parent, children] of categories) {
    for (let subCategory of children) {
```
Here, `typeof parent` is `Category`:
image


Continues the [line 156](https://github.com/jdneo/vscode-leetcode/blob/d9abd5f68a92bc7679066ea35bd342fd80a599ea/src/leetCodeExplorer.ts#L156):
```typescript
const problems: list.IProblem[] | undefined = this.treeData[parent].get(subCategory);
```

Since `typeof parent` is `Category` and `this.treeData`'s key type is `string`, `get`'s type could not be deduced from the usage, and there's no way to use `Category.xxxx` here:
image

Meanwhile, we couldn't change `this.treeData` type to `Map<Category, Map<string, list.IProblem[]>> since there is a field `Favorite` which doesn't processed as `Category` and its value type is `list.IProblem[]`.

If you insist to use enum here, a workaround could be:
```typescript
private treeData: {
    [Category.Difficulty]: Map<string, list.IProblem[]>,
    [Category.Tag]: Map<string, list.IProblem[]>,
    [Category.Company]: Map<string, list.IProblem[]>,
    Favorite: list.IProblem[],
};
```

Now `get`'s type could be deduced properly:
image

Yet I think it complicates what could have been an easy and elegant solution in Typescript.

jdneo:
Merged with some refactoring

src/leetCodeExplorer.ts
jdneo: We need to write more hard-coded strings if using string union here while using enum can get rid of this. That's the reason I prefer using enum here: make the code easier to maintain in the future.

There are many ways to get avoid the special field `Favorite`. One simple way to do a simple check before getting the map. TS will help us to infer the type during compilation. See my code for more details.

PR #114, opened 2019-02-10
add switchDefaultLanguage

ZhangYet:
Once you choose the default language,  every time you show the new problem with the default language. If you want to change another language, you need to change it in Preference. I think it's convenient for a command to switch default language.


jdneo:
Hi @ZhangYet 

Thank you for your contribution.

May I ask the background about this PR first? Because I want to better understand how users are using this extension.

Does that mean each time you want to practice a problem, you would like to practice it with multiple languages? How frequently will the 'switching default language' happen?

ZhangYet:
I use Go and Python in my job, so I practice leetcode with go and python. And now I'm going to learn rust so I would like to answer problems with rust.

However, I'm not sure if anyone else wants to switch their default language frequently.

jdneo:
@ZhangYet Got it.

So from my understanding, you are used to practicing one language for a period, and maybe another language later. During a period of time, you still prefer using one default language rather than selecting the language explicitly each time when picking a problem. Am I right?

ZhangYet:
@jdneo In fact, I will switch to python if I have trouble when using other languages.
Anyway if you think it's not a practical feature, you can close this PR.  I agree that maybe few people would use this feature

jdneo:
@ZhangYet Hmm, Actually I do not want to close this PR for now, cuz we do not know if other users need this feature as well or not. Instead, I create a tracking issue #115. If any user wants this, they can click the 👍 to let us know.

Back to the problem you meet now, the workaround is to remove the default language setting, then the extension will ask you to select the language each time you pick a problem and click something like `Never show again` when the extension asks you to set the default language.

At last, I still want to thank you for sending out this PR. 😄

ZhangYet:
@jdneo LGTM tks

jdneo:
@ZhangYet There are several users click  👍 in the related issue. I think we can start working on this feature now.

src/commands/plugin.ts
```diff
@@ -4,10 +4,19 @@
 import * as vscode from "vscode";
 import { leetCodeExecutor } from "../leetCodeExecutor";
 import { IQuickItemEx } from "../shared";
-import { Endpoint } from "../shared";
+import { Endpoint, languages } from "../shared";
 import { DialogType, promptForOpenOutputChannel, promptForSignIn } from "../utils/uiUtils";
 import { deleteCache } from "./cache";
 
+export async function switchDefaultLanguage(): Promise<void> {
+    const leetCodeConfig: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("leetcode");
```
jdneo: The code here is somehow duplicated with https://github.com/jdneo/vscode-leetcode/blob/master/src/commands/show.ts#L44-L51

Can we extract them out to share only one piece of code?

src/commands/plugin.ts
```diff
@@ -4,10 +4,19 @@
...
+    const language: string | undefined = await vscode.window.showQuickPick(languages, { placeHolder: "Select the language you want to use" });
```
jdneo: If the user has already set a default language, we should at least add a ✅mark before the language.

jdneo:
@ZhangYet Would you like to keep working on this PR?

ZhangYet:
@jdneo Fine. I will fix the conflict.

src/commands/plugin.ts
ZhangYet: I will do it later for I am going to change my job these days.

jdneo:
Fixed in #178 

@ZhangYet Still want to thank you for the engagement and the inspiring feature request 👍

Issue #119, opened 2019-02-12
Customize the sub-directory to save the file generated by 'Show Problem'


liuyangc3:
例如把生成文件的目录设置为, 当前打开文件夹下的 任意某个文件夹

jdneo:
Is #120 what you are asking for?

liuyangc3:
@jdneo not quite same,for example is it possiable generate `.java` file in `somedir`

![image](https://user-images.githubusercontent.com/6705844/52688563-f8a77c00-2f91-11e9-934e-7bc00602b2bf.png)

ringcrl:
Maybe should add a vscode config like：

"leetcode.outputPath": "tag", // somedir Or other

jdneo:
@ringcrl I'm thinking exactly the same thing.

We need to figure out if it is possible to combine this issue with #120 and give out an overall solution.

ringcrl:
| Setting Name | Description | Default Value |
|---|---|---|
| `leetcode.outputPath` | Which path will the problem generate to. `root` for  root directory，`tag` for create a directory based on the tag of the problem，`custom_path` for User-defined directory  | `root` |

jdneo:
@ringcrl LGTM, just some small improvement.

The `leetcode.outputPath` specifies the relative path to save the problem files. Meanwhile, we have several preserved values for it, which will be resolved according to the actual situation. Those preserved values are:
- `${tag}` - a directory based on the problem's tag. For example, if the problem belongs to: `Binary Indexed Tree`, then a folder named `binary_indexed_tree` will be used here.
- `${language}` - a directory based on the language used. For example, `java` for Java language
- `${difficulty}` - a directory based on the problem's difficulty. For example, `easy`
> If this setting is not set, the files will be generated to `root`

ringcrl:
OK~I will finish it later

jdneo:
@ringcrl Great! Thank you.

jdneo:
@all The PR has been merged and will be released in 0.11.0.

Thank you all for your participation.

wangjunwei87:
I really need this feature, when will version 0.11.0 release?

jdneo:
@wangjunwei87 It should be released recently

jdneo:
@wangjunwei87 0.11.0 has been released

wangjunwei87:
@jdneo Thanks for your great work, and I am wondering if it's possible to use preserved value for interpolation, like "something-${tag}-${language}-something".

[...]

PR #123, opened 2019-02-13
add outputPath Config

ringcrl:
add outputPath config for leetcode #119

Change Request,Need Help...Find a way to get a Problem difficulty...

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,17 +114,16 @@
 
 
 ## Settings
-| Setting Name               | Description                                                                                                                                                                                                 | Default Value |
-| -------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------- |
-| `leetcode.hideSolved`      | Specify to hide the solved problems or not                                                                                                                                                                  | `false`       |
-| `leetcode.showLocked`      | Specify to show the locked problems or not. Only Premium users could open the locked problems                                                                                                               | `false`       |
-| `leetcode.defaultLanguage` | Specify the default language used to solve the problem. Supported languages are: `bash`, `c`, `cpp`, `csharp`, `golang`, `java`, `javascript`, `kotlin`, `mysql`, `python`,`python3`,`ruby`,`scala`,`swift` | `N/A`         |
-| `leetcode.useWsl`          | Specify whether to use WSL or not                                                                                                                                                                           | `false`       |
-| `leetcode.endpoint`        | Specify the active endpoint. Supported endpoints are: `leetcode`, `leetcode-cn`                                                                                                                             | `leetcode`    |
-| `leetcode.outputPath`      | * `${tag}` - a directory based on the problem's tag. For example, if the problem belongs to: `Binary Indexed Tree`, then a folder named `binary_indexed_tree` will be used here.                            |
-* `${language}` - a directory based on the language used. For example, `java` for Java language
-* `${difficulty}` - a directory based on the problem's difficulty. For example, `easy`
-* If this setting is not set, the files will be generated to the base path of the workspace folder                                                                                                                     | `root`        |
+| Setting Name | Description | Default Value |
+|---|---|---|
+| `leetcode.hideSolved` | Specify to hide the solved problems or not | `false` |
+| `leetcode.showLocked` | Specify to show the locked problems or not. Only Premium users could open the locked problems | `false` |
+| `leetcode.defaultLanguage` | Specify the default language used to solve the problem. Supported languages are: `bash`, `c`, `cpp`, `csharp`, `golang`, `java`, `javascript`, `kotlin`, `mysql`, `python`,`python3`,`ruby`,`scala`,`swift` | `N/A` |
+| `leetcode.useWsl` | Specify whether to use WSL or not | `false` |
+| `leetcode.endpoint` | Specify the active endpoint. Supported endpoints are: `leetcode`, `leetcode-cn` | `leetcode` |
+
+## Troubleshooting
+When you meet any problem, you can check the [Troubleshooting page](https://github.com/jdneo/vscode-leetcode/wiki/Troubleshooting) first.
 
 ## Release Notes
 
--- a/README.md
+++ b/README.md
@@ -67,8 +67,7 @@
   <img src="https://raw.githubusercontent.com/jdneo/vscode-leetcode/master/docs/imgs/pick_problem.png" alt="Pick a Problem" />
 </p>
 
-- Right click the problem in the `LeetCode Explorer` and select `Show problem` will generate a new file with the problem description for you.
-- Right click the problem in the `LeetCode Explorer` and select `Show problem with tag` will generate a new file with the problem description and classified by tag folder.
+- Right click the problem in the `LeetCode Explorer` and select `Show Problem` will generate a new file with the problem description for you.
 
   > Note: If no folder is opened in VS Code, the extension will save the problem files in **$HOME/.leetcode/**.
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -114,13 +114,17 @@
...
 
 ## Settings
-| Setting Name | Description | Default Value |
-|---|---|---|
-| `leetcode.hideSolved` | Specify to hide the solved problems or not | `false` |
-| `leetcode.showLocked` | Specify to show the locked problems or not. Only Premium users could open the locked problems | `false` |
-| `leetcode.defaultLanguage` | Specify the default language used to solve the problem. Supported languages are: `bash`, `c`, `cpp`, `csharp`, `golang`, `java`, `javascript`, `kotlin`, `mysql`, `python`,`python3`,`ruby`,`scala`,`swift` | `N/A` |
-| `leetcode.useWsl` | Specify whether to use WSL or not | `false` |
-| `leetcode.endpoint` | Specify the active endpoint. Supported endpoints are: `leetcode`, `leetcode-cn` | `leetcode` |
+| Setting Name               | Description                                                                                                                                                                                                 | Default Value |
```
jdneo: Hmm, please revert the changes to the markdown document. Let me do this.

jdneo:
@ringcrl 

I think there is still some misunderstanding here. The aim of add the setting `leetcode.outputPath` is to make it possible for the users to categorize the files as their wish. So instead of make it as a enum setting, we should make it as a plain string.

Let me explain more about it:

The setting of `leetcode.outputPath` has no default value, which means it is empty by default. And, by default, when the user click `show problem`, the file goes to the base path of the workspace folder.

If the user set some values to it, for example: 'happy' then the file goes to the folder called `happy`

Meanwhile, besides let the user customize the output folder, we will also provide some **special** values, which are, as you know, `${tag}`, `${language}`, `${difficulty}`. This means the user needs to explicitly brace the value with `${}`, then the extension will get to know that: Oh, this is not a general path. I need to traslate it to another value...

Do you get my point?

src/commands/show.ts
```diff
@@ -49,7 +49,25 @@ async function showProblemInternal(id: string): Promise<void> {
...
 
-        const outDir: string = await selectWorkspaceFolder();
+        let outDir: string = await selectWorkspaceFolder();
+        const outputPath: string = leetCodeConfig.get<string>("outputPath") || "root";
+        switch (outputPath) {
+            case "root": {
+                break;
+            }
+            case "tag": {
+                const { tags } = await leetCodeExecutor.getCompaniesAndTags();
```
jdneo: `Tag` and `Difficulty` is stored in the `LeetCodeNode`. 

See: https://github.com/jdneo/vscode-leetcode/blob/master/src/explorer/LeetCodeNode.ts#L28-L34
https://github.com/jdneo/vscode-leetcode/blob/master/src/commands/show.ts#L15-L19

ringcrl:
Thank you very much, I have made changes.

src/commands/show.ts
```diff
@@ -49,7 +50,34 @@ async function showProblemInternal(id: string): Promise<void> {
             return;
         }
 
-        const outDir: string = await selectWorkspaceFolder();
+        let outDir: string = await selectWorkspaceFolder();
+        const outputPath: string = leetCodeConfig.get<string>("outputPath") || "";
+        const problem: IProblem | undefined = listProblems.find((item: IProblem) => item.id === id);
+        switch (outputPath) {
```
jdneo: Steps to treat with the `outputPath`:

## Step 1: test if it's braced by `${}`:
> You can do this using regex: `\$\{(.*?)\}`, the regex can help you extract the value inside the braces

If yes, translate for `tag`, `language`, `difficulty`

### Why?
1. If the user provides a value which is not supported yet, for example: `${company}`, we will have a chance to tell the user this category is not supported yet
2. Use `switch(value.toLocaleLowerCase())` to make sure if the user provides `${Difficulty}`, the feature will still work.

## Step 2: join the path
If it's not braced by `${}`, then we can directly join the path without translation.

There is no need to treat with empty value because:

1. `leetCodeConfig.get<string>("outputPath") || ""` make sure the value will not be undefined
2. `path.join("a", "")` will ignore the empty string

ringcrl:
Thank you very much for these helpful suggestions.

src/commands/show.ts
```diff
@@ -34,10 +36,10 @@ export async function searchProblem(): Promise<void> {
     if (!choice) {
         return;
     }
-    await showProblemInternal(choice.value);
+    await showProblemInternal(problems.find((problem: IProblem) => problem.id === choice.value) as IProblem);
```
jdneo: Have you tried to change the line 118 from 
`value: problem.id,` to `value: problem`?

src/commands/show.ts
```diff
@@ -16,16 +17,17 @@ export async function showProblem(node?: LeetCodeNode): Promise<void> {
     if (!node) {
         return;
     }
-    await showProblemInternal(node.id);
+    await showProblemInternal(node);
```
jdneo: Though type casting from `LeetCodeNode` to `IProblem` is fine. Let's add getter in `LeetCodeNode` to get the data, which is typed as `IProblem`.

> It looks like I need to do some refactors here later after this PR, The definition of `LeetCodeNode` and `IProblem` is overlapped.

src/commands/show.ts
```diff
@@ -49,9 +51,36 @@ async function showProblemInternal(id: string): Promise<void> {
...
         }
 
-        const outDir: string = await selectWorkspaceFolder();
+        let outDir: string = await selectWorkspaceFolder();
+        const outputPathCfg: string = leetCodeConfig.get<string>("outputPath") || "";
+        const outputPath: RegExpMatchArray | null = outputPathCfg.match(/\$\{(.*?)\}/);
+        if (outputPath) {
+            switch (outputPath[1].toLowerCase()) {
+                case "tag": {
+                    const closestTag: string = node.tags.reduce((prev: string, curr: string) => {
```
jdneo: Hmm, here is the problem, looks like you want to find out the tag with the longest length if the problem has multiple tags. Any reason for this logic?

src/commands/show.ts
```diff
@@ -49,9 +51,36 @@ async function showProblemInternal(id: string): Promise<void> {
...
+                            prev;
+                    }, "");
+                    outDir = path.join(outDir, closestTag);
+                    break;
+                }
+                case "language": {
+                    outDir = path.join(outDir, language);
+                    break;
+                }
+                case "difficulty": {
```
jdneo: No need to add `{}` for each `case` and `default`

src/commands/show.ts
ringcrl: I found that if I took the first one, most of them are [array]. This is not what I want. By observing the all of tags, I found that the longest tag is generally the best one...I originally wanted to add a priority to the tag, which is complicated.

src/commands/show.ts
ringcrl: emm...Because of type error, so I use IProblem on both sides.

src/commands/show.ts
jdneo: Then, this is not convincing enough. Please considering:
- We have no idea what does user want
- There is no data to support the conclusion: `the longest tag is generally the best one`, it's just a subjective judgment

Meanwhile, by doing this, we provided an implicit logic for the user, this is not a recommended user experience.

I have two suggestions:
1. show quick picks to the user if multiple tags are available, let the user choose one.
2. Remove the support for `${tag}` for now.

src/commands/show.ts
jdneo: I mean, you are using `problems.find((problem: IProblem) => problem.id === choice.value)` to get the problem, that is because the `choice.value` only stores the `id`. You can make it to store the problem itself.

src/commands/show.ts
ringcrl: Thanks for your suggestion，This solution is very reasonable

ringcrl:
Add tag selector

jdneo:
@ringcrl Looks much better now.

I just did some changes and merged the PR.

Thank you for your contribution.

PR #143, opened 2019-02-20
Add tooltips for sub-category folder

pujiaxun:
Background: I wonder how many problems I have solved. Then I found the tooltip field is available for now. That's it~

### DIFF
--- a/src/explorer/LeetCodeTreeDataProvider.ts
+++ b/src/explorer/LeetCodeTreeDataProvider.ts
@@ -46,6 +46,7 @@ export class LeetCodeTreeDataProvider implements vscode.TreeDataProvider<LeetCod
         const idPrefix: number = Date.now();
         return {
             label: element.isProblem ? `[${element.id}] ${element.name}` : element.name,
+            tooltip: this.getSubCategoryTooltip(element),
             id: `${idPrefix}.${element.parentName}.${element.id}`,
             collapsibleState: element.isProblem ? vscode.TreeItemCollapsibleState.None : vscode.TreeItemCollapsibleState.Collapsed,
             contextValue: element.isProblem ? "problem" : element.id.toLowerCase(),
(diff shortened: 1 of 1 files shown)


jdneo:
@pujiaxun Sounds great. Would you mind to attach a screenshot in the PR?

pujiaxun:
> @pujiaxun Sounds great. Would you mind to attach a screenshot in the PR?

image

It looks like this. @jdneo

jdneo:
@pujiaxun Just one small comment.

pujiaxun:
```js
        import * as os from "os";
        //...

        return [
            `AC: ${numAC}`,
            `Not AC: ${numNotAC}`,
            `Unknown: ${numUnknown}`,
            `Total: ${problems.length}`,
        ].join(os.EOL);
```

Like this? @jdneo

jdneo:
@pujiaxun Yes, looks much better.

pujiaxun:
@jdneo  Any other suggestions? For instance, I think the name "Not AC" sounds weird...

jdneo:
@pujiaxun hmm, good question.

Regarding the wording. My suggestion is `Not AC` -> `Failed`

Either `Unknown` or `Total`, one of them can be removed I think? What do you think?

pujiaxun:
@jdneo I prefer to keep `Total` if I have to give up one of them. After all, the tag `Unknown` is somewhat confusing.

jdneo:
Same to me. So let's keep `AC`, `Failed` and `Total`.

src/explorer/LeetCodeTreeDataProvider.ts
```diff
@@ -180,7 +180,6 @@ export class LeetCodeTreeDataProvider implements vscode.TreeDataProvider<LeetCod
 
         let numAC: number = 0;
         let numNotAC: number = 0;
```
jdneo: nit:
`numAC` -> `acceptedNum`
`numNotAC` -> `failedNum`

pujiaxun: cool

Issue #117, opened 2019-02-12
proxy issue: log in time out


shaoxul:
Dear friends,

For the sake of the gateway firewall, the http proxy must be used to connect internet.
But the log-in always times out. It looks like to try to connect to the leetcode server directly instead of through the proxy.

Is there any configuration I missed?

jdneo:
@shaoxul Hmm, I have not met this problem before. But just found that the VS Code has proxy related settings. Maybe you can have a try?

![image](https://user-images.githubusercontent.com/6193897/52612638-a39e3400-2ec5-11e9-84d0-18e9b0a41036.png)

shaoxul:
@jdneo Thank you for your reply. Yes, I tried, but it works for the VS code itself(e.g. extension market), but not works for this this extension.

jdneo:
So far, I have no idea what I can do for your problem...

Need some more investigation here...

MiloJason:
@shaoxul I have met the same problem as you.

zfrxiaxia:
same problem

edvardchen:
@shaoxul  It seems that vscode's proxy setting does't work. Try to start vscode with command line arguments

```
code --proxy-server=yourproxy
``` 

or setting the global env variables before run code. [See more in official document ](https://code.visualstudio.com/docs/setup/network)

jdneo:
@edvardchen Provided a solution in #147.

The root cause is that when we trigger a leetcode-cli command, we will spawn a new child_process to send the request, which may bypass the proxy setting in VS Code.

So, we need to read the proxy setting of VS Code. and set it in the created child_process.

More details can be found in #147

PR #147, opened 2019-02-21
Pass http proxy as env to command (fix #136, #117)

### LINKED ISSUE #136: mac login error
login: pass: - Signing in leetcode.com
[ERROR] Error: tunneling socket could not be established, cause=connect EHOSTUNREACH 0.0.4.63:80 - Local (10.0.0.11:59146)

edvardchen:
We use leetcode-cli and leetcode-cli uses request to call LeetCode web API.
It's lucky that request respects the http proxy env variables(http_proxy, https_proxy),
so all we need to do is passing proxy as env variable http_proxy to every command.

TODO: do we need to set https_proxy for https request?

### DIFF
--- a/src/utils/workspaceUtils.ts
+++ b/src/utils/workspaceUtils.ts
@@ -44,18 +44,3 @@ export async function getActiveFilePath(uri?: vscode.Uri): Promise<string | unde
 export function getWorkspaceConfiguration(): vscode.WorkspaceConfiguration {
     return vscode.workspace.getConfiguration("leetcode");
 }
-
-function getHttpAgent(): string | undefined {
-    return vscode.workspace.getConfiguration("http").get<string>("proxy");
-}
-
-// clone process.env and add http proxy
-export function createEnvOption(): {} {
-    const proxy: string | undefined = getHttpAgent();
-    if (proxy) {
-        const env: any = Object.create(process.env);
-        env.http_proxy = proxy;
-        return env;
-    }
-    return process.env;
-}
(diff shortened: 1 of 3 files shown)


jdneo:
Thank you @edvardchen for contribution.

So Is this PR aimed to fix #117

edvardchen:
Yes.

jdneo:
@edvardchen Awesome, I'll take a look later, thanks!

jdneo:
Hi @edvardchen

What will happen if we don't set https_proxy for https request

edvardchen:
request will use http_proxy as fallback

edvardchen:
Because vscode user setting doesn't have a https.proxy option, so we should not worry about it.

src/utils/workspaceUtils.ts
```diff
@@ -44,3 +44,18 @@ export async function getActiveFilePath(uri?: vscode.Uri): Promise<string | unde
 export function getWorkspaceConfiguration(): vscode.WorkspaceConfiguration {
     return vscode.workspace.getConfiguration("leetcode");
 }
+
+function getHttpAgent(): string | undefined {
+    return vscode.workspace.getConfiguration("http").get<string>("proxy");
+}
+
+// clone process.env and add http proxy
+export function createEnvOption(): {} {
```
jdneo: Looks like cpUtils.ts is a better place to put this method?

edvardchen: Sure. I'll move it there.

jdneo:
@edvardchen Thank you for your contribution!

Issue #59, opened 2018-08-15
如何在show problem时，读取本地文件，而不是新建一个？


SkyLin0909:
![image](https://user-images.githubusercontent.com/20888672/44142674-816bef8a-a0b3-11e8-99be-05f386798e7f.png)

jdneo:
不太确定我有没有正确理解的你的问题。

你可以试试在VS Code里，点击`File` > `Open Folder...`， 然后选择目标文件夹（默认情况下所有题目会存储在`$HOME/.leetcode/`），这个时候你就可以在VS Code里看到所有文件了。

![image](https://user-images.githubusercontent.com/6193897/44145903-dd08c87a-a0bf-11e8-871d-db2c14ad1ee8.png)

taoweicn:
> 不太确定我有没有正确理解的你的问题。
> 
> 你可以试试在VS Code里，点击`File` > `Open Folder...`， 然后选择目标文件夹（默认情况下所有题目会存储在`$HOME/.leetcode/`），这个时候你就可以在VS Code里看到所有文件了。
> 
> ![image](https://user-images.githubusercontent.com/6193897/44145903-dd08c87a-a0bf-11e8-871d-db2c14ad1ee8.png)

我觉得在点击show problem时先判断一下本地有没有文件再来决定打开本地文件（如果本地有）或者拉取题目（如果本地没有）这样好一点，因为这样可以看到之前的草稿或者提交的答案

jdneo:
@taoweicn This is a good suggestion.

@SkyLin0909 Do you mean to make the extension work as @taoweicn said?

SkyLin0909:
@jdneo yes

jdneo:
@poppinlp All the issues are PR welcomed. Just remember to discuss the implementation in the issue before sending out the PR.

Thanks

poppinlp:
@jdneo 
Thanks. Nice!
I've looked into the code about this issue since it really bothers me.

This issue is caused by the params we passed into the `show` API of `leetcode-cli` which will generate a new source code file every time. And i haven't seen a param for existence checking in `show` API.

So here may be some plans:

A: We could genearte the file name by ourself and by which we could do the existence checking to determine whether to use the local code or get the remote template by `show` API from `leetcode-cli`.

This could solve this problem but we may meet some new problems too:

- generating file name by id and slug which should be compatible with the present
- mapping the file extension with language

[...]

PR #149, opened 2019-02-22
feat: support load code from local file for issue 59

poppinlp:
This RP is for #59. The main logic change is that `showProblem` operation will try to load local code first. Here are the changes:

- Add `lodash.kebabcase` in dependencies to generate a slug
- Add `langExt` in `shared.ts` to do the extension mapping for language
- Add `problemUtils.ts` which contains utils for problem
- The `showProblem` in `leetCodeExecutor.ts` will return the code file path right now

### DIFF
--- a/src/leetCodeExecutor.ts
+++ b/src/leetCodeExecutor.ts
@@ -8,7 +8,7 @@ import * as requireFromString from "require-from-string";
 import * as vscode from "vscode";
 import { Endpoint, IProblem } from "./shared";
 import { executeCommand, executeCommandWithProgress } from "./utils/cpUtils";
-import { genFileName } from './utils/problemUtils';
+import { genFileName } from "./utils/problemUtils";
 import { DialogOptions, openUrl } from "./utils/uiUtils";
 import * as wsl from "./utils/wslUtils";
 
@@ -78,9 +78,8 @@ class LeetCodeExecutor {
     public async showProblem(node: IProblem, language: string, outDir: string): Promise<string> {
         const fileName: string = genFileName(node, language);
         const filePath: string = path.join(outDir, fileName);
-        const hasLocalCode: boolean = await fse.pathExists(filePath);
 
-        if (!hasLocalCode) {
+        if (!await fse.pathExists(filePath)) {
             const codeTemplate: string = await this.executeCommandWithProgressEx("Fetching problem data...", "node", [await this.getLeetCodeBinaryPath(), "show", node.id, "-cx", "-l", language]);
             await fse.writeFile(filePath, codeTemplate);
         }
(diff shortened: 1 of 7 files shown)


src/utils/problemUtils.ts
```diff
@@ -0,0 +1,8 @@
+import kebabCase = require("lodash.kebabcase");
+import { IProblem, langExt } from "../shared";
+
+export function genFileName(node: IProblem, language: string): string {
+    const name: string = kebabCase(node.name);
+    const ext: string | undefined = langExt.get(language);
+    return `${node.id}.${name}.${ext}`;
```
jdneo: `ext` here might be `undefined`. Though according to current logic, it's impossible.

I think you can have another util method here to parse the `ext` from a language name. Meanwhile, throw an error if the language is not supported.

src/utils/problemUtils.ts
```diff
@@ -0,0 +1,8 @@
+import kebabCase = require("lodash.kebabcase");
```
jdneo: Use `import` instead of `require`.

src/leetCodeExecutor.ts
```diff
@@ -74,8 +75,17 @@ class LeetCodeExecutor {
         );
     }
 
-    public async showProblem(id: string, language: string, outDir: string): Promise<string> {
-        return await this.executeCommandWithProgressEx("Fetching problem data...", "node", [await this.getLeetCodeBinaryPath(), "show", id, "-gx", "-l", language, "-o", `"${outDir}"`]);
+    public async showProblem(node: IProblem, language: string, outDir: string): Promise<string> {
+        const fileName: string = genFileName(node, language);
+        const filePath: string = path.join(outDir, fileName);
+        const hasLocalCode: boolean = await fse.pathExists(filePath);
```
jdneo: `hasLocalCode` is redundant. We can move `await fse.pathExists(filePath)` into the if block.

src/utils/problemUtils.ts
poppinlp: I use it in this way because of the issue of lodash. T_T

src/utils/problemUtils.ts
poppinlp: OK. I'll fix it.

src/leetCodeExecutor.ts
poppinlp: OK.

src/utils/problemUtils.ts
jdneo: OK, it's fine.

poppinlp:
I'm sorry for the lint error. T_T

jdneo:
@poppinlp Thank you for the contribution. Nice job!

I'll send out another PR to add you the contributors..

Issue #166, opened 2019-02-28
Custom file name template


Dec-F:
考虑支持自定义的文件名吗？
 
比如 中文版生成的文件名含有中文，import一个中文文件，感觉不是很靠谱，所以每次都要手动修改。

我觉得，文件名最好支持模版插值，可以取{tag}{difficulty}等几个meta信息填充，这样也可以满足不同人的需求。

jdneo:
@poppinlp Guessing this might be a regression in #149

If the user is using Chinese version LeetCode, the file name we generated will contain Chinese

Dec-F:
> @poppinlp Guessing this might be a regression in #149
> 
> If the user is using Chinese version LeetCode, the file name we generated will contain Chinese

I think these two issues are not the same function, maybe you didn't understand what I meant. I am suggesting to add a configuration item, just like `outputFolder`, does not affect the reading of local files.

Dec-F:
Setting Name | Description | Default Value
------------ | ------------- | ------------
fileName | The name of the generated file;{id}-{difficulty}-{tag}-{name} | {id}.{name}.{ext} |


like this? @jdneo

poppinlp:
@jdneo 
I have not tested on the Chinese version LeetCode. That's my fault. I'll fix it and think about the test cases.

@Dec-F 
Yep. You are talking about a different feature. I think I get your point.
What's your option about this? @jdneo I think this feature is reasonable but not high priority.

Dec-F:
If this feature is necessary, I will submit a PR tomorrow or weekend. What do you think? @jdneo @poppinlp

jdneo:
@poppinlp Not your fault, we all missed that part, It's great if this can be fixed.

@Dec-F PR is welcome, you can send out it whenever you want.

poppinlp:
@jdneo  I've sent a PR to fix the Chinese file name issue.

@jdneo @Dec-F Here maybe a plan to support the file name pattern:

We could pass the whole config into the `genFileName` function instead of the only `endpoint` right now to avoid too many params. And then generate the file name by the pattern in the setting. And don't forget to update the documents after that.

The variable in pattern could be:
- id
- name
- difficulty
- ext
- companies(join the array in kebabcase?)
- tags(join the array in kebabcase?)

jdneo:
for a certain problem, it may belong to multiple companies and tags, how to handle this senario?

poppinlp:
[...]

PR #168, opened 2019-02-28
fix: use pinyin instead of Chinese in file name for leetcode-cn problem

poppinlp:
This PR is for #166 

- Add a dependency [pinyin](https://github.com/hotoo/pinyin) to generate slug for Chinese.
- Add a function to generate slug by problem name and endpoint.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,9 +5,9 @@
     "requires": true,
     "dependencies": {
         "@types/fs-extra": {
-            "version": "5.0.0",
-            "resolved": "https://registry.npmjs.org/@types/fs-extra/-/fs-extra-5.0.0.tgz",
-            "integrity": "sha512-qtxDULQKUenuaDLW003CgC+0T0eiAfH3BrH+vSt87GLzbz5EZ6Ox6mv9rMttvhDOatbb9nYh0E1m7ydoYwUrAg==",
+            "version": "5.0.5",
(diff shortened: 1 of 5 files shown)


jdneo:
@poppinlp Thank you for providing the fix.

Hmm, what I'm actually thinking is that can we make sure that using `Pinyin` is better than `Chinese` for our users. At least for me, I have no answers to this question now.

So my suggestion is we can just leave this fix for a while, wait for our users' feedback and decide where we should go, what do you think?

poppinlp:
For me, I think it's a little bit strange to use Chinese in a file name.
But I agree with you that we could wait for more feedback. :)

magic-akari:
Can we fetch the meta information from leetcode and use it in leetcode-cn?

jdneo:
> Can we fetch the meta information from leetcode and use it in leetcode-cn?

Do you mean make the problem from `leetcode-cn` has the same name as those from `leetcode`, instead of using Chinese name?

magic-akari:
> > Can we fetch the meta information from leetcode and use it in leetcode-cn?
> 
> Do you mean make the problem from `leetcode-cn` has the same name as those from `leetcode`, instead of using Chinese name?

Yes, that is what I mean.

jdneo:
Cool, That seems a better solution. Do you want to have a look into this?

poppinlp:
Seems this PR is outdated with lots of conflicts. We may create another one for that solution.

PR #174, opened 2019-03-01
✨ add a new config:'fileName'

Dec-F:


jdneo:
@Dec-F Thank you for contribution. While, the PR contains too much formatting change, which makes it difficult to review.

Another question is that will your request get satisfied if we implement the comment here: https://github.com/jdneo/vscode-leetcode/issues/119#issuecomment-466778120

jdneo:
@Dec-F Oh, I missed one thing about this issue.

The leetcode-cli has a rule on the file name otherwise the submission will fail. So this PR might be hanged for now, until the dependency on leetcode-cli removed.

Sorry for that I forgot this limitation at first.

Dec-F:
@jdneo Ok, then what are the plans for lifting this dependency? I can see what I can do.

jdneo:
@Dec-F yes, we do need some more people to write the core lib. A lot of work to do...

I'm still thinking about how to achieve that and where to start...

Will inform you when I have an idea.

magic-akari:
Hello. Any update here?

jdneo:
Sorry I do not update in time about this PR. The limitation for the file name is not valid anymore. Since the generated file will contains a comment to contain problem metadata, such as the problem id.

jdneo:
Hi @Dec-F,

Thank you for your contribution and great work! We now have a new setting: [`leetcode.filePath`](https://github.com/jdneo/vscode-leetcode/wiki/Customize-the-Relative-Folder-and-the-File-Name-of-the-Problem-File) which I think can resolve this feature request.

I'm closing this PR but feel free to let me know if you have more ideas on this.

Thanks!

Issue #165, opened 2019-02-28
The icon in viewsContainers is blur in MacOSX


Eskibear:
You might compare it with other icons, it looks uncomfortable. 

<img width="51" alt="screen shot 2019-02-28 at 1 18 12 pm" src="https://user-images.githubusercontent.com/2351748/53542860-58d61a80-3b5b-11e9-911c-0db43715a681.png">

jdneo:
Fixed in #175 

Thank you @poppinlp for help!

PR #175, opened 2019-03-01
fix: use a new icon svg

poppinlp:
This PR is for #165 
Just use a new SVG file.

Here's a screenshot on my retina screen:
image

Here's a screenshot on my non-retina screen:
image

Seems it will fix the blur on a retina screen and won't break the non-retina screen.


jdneo:
@poppinlp

Genius!! how did you achieve that?

Issue #105, opened 2019-02-06
Cannot retrieve time and memory result


dumplingslove:
√ Accepted
  √ 258/258 cases passed (88 ms)
[WARN] Failed to get submission beat ratio.

jdneo:
@dumplingslove What's the endpoint are you using? `leetcode` or `leetcode-cn`? Do you still have the issue on your side?

jdneo:
Related issue: https://github.com/skygragon/leetcode-cli/issues/156

dumplingslove:
Thanks for the checking.

BrambleXu:
I have the same issue. Hope to see the new version to fix it.

zfrxiaxia:
same problem, leetcode-cn

jdneo:
Resolved in #179.

Issue #103, opened 2019-02-04
Add PHP support


sohaha:
目前 leetcode 是支持 php 的了，能否考虑加上？
（不加其实也没事，估计也没 写 php 的会用吧🙅‍♂️，逃）

poppinlp:
@sohaha Hi
This feature depends on the leetcode-cli package.
I've sent a [PR](https://github.com/skygragon/leetcode-cli/pull/163) for PHP supporting and we'll support it after they merge and release. :)

jdneo:
Considering the leetcode-cli is inactive for a long time. I'm thinking to fork and publish a temporary bit for it. Because currently there are too many issues blocked.

poppinlp:
@jdneo And maybe we should export a JS module instead of the CLI for easy using and performance reason.

jdneo:
@poppinlp That's exactly what I'm thinking.

But exporting a JS module might cost much more time, so in the short term, publishing a forked cli bit could be a workaround. For the long term, we need a library instead of depending on the cli.

poppinlp:
@jdneo Agree. 👍

And if you like, maybe we could create a github org to contains the JS module, the CLI based on JS module, and extensions.
And i just find out that the owner of [leetcode package name](https://www.npmjs.com/package/leetcode) is a friend of me and it's long term inactive. So maybe we could use that package name. :)

jdneo:
@poppinlp Ok, let me think more about it and I'll let you know the update when I have more thoughts on this. Thank you for your passion.

jdneo:
Resolved in #179.

Issue #83, opened 2019-01-03
Add rust support?


Airtnp:
https://leetcode.com/discuss/general-discussion/136640/Rust-support/208982

jdneo:
Rust has been added in the leetcode-cli but the new version has not been released.

Will add it when it gets released.

jdneo:
Resolved in #179.

PR #179, opened 2019-03-03
Use internal build cli

### LINKED ISSUE #74: Power operator displays in a wrong way
pow will change
i think you could instend `^`

jdneo:
Try to resolve #74, #83, #103, #105

@poppinlp Please help review when you have time. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@
 |---|---|---|
 | `leetcode.hideSolved` | Specify to hide the solved problems or not | `false` |
 | `leetcode.showLocked` | Specify to show the locked problems or not. Only Premium users could open the locked problems | `false` |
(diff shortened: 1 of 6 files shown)


jdneo:
Let me explain more about this PR.

In this PR, we change the dependency from `leetcode-cli` to `vsc-leetcode-cli`, which is a forked privacy build of the `leetcode-cli`. This is due to that the latest version of `leetcode-cli` has some broken features and inactive for a long time. So I have to replace it with my forked version.

So far I have no plan to write a new cli tool to replace it. I'm very willing to merge back to the origin `leetcode-cli` when it becomes active again.

taoweicn:
Using a forked version of `leetcode-cli` is a better choice I think, we can keep in sync with the upstream(`leetcode-cli`), and then fix known issues ASAP or customize the extra features we need.

Issue #132, opened 2019-02-17
Click the tree explorer should open the topic


yaohaizh:
A.T.T

jdneo:
Resolved in #186

PR #186, opened 2019-03-05
Add Click the tree explorer should open the topic (#132)

### LINKED ISSUE #131: Feature request - Better UI to display the question
Current, the topic/question is in a single file with comments, which is not user friendly to read and write code. 

Should leverage the web view to provide some side by side display experience. Or user can directly display the leetcode.com page?

xuzaixian:
Resolve #131


jdneo:
@xuzaixian Thank you for contribution.

From my understanding, after applying this change, when the user clicks on the tree node, we will generate new / open existing problem file, right?

xuzaixian:
yes ~

jdneo:
Emm, I'm ok to open the existing problem file.

While for the non-existing file, it will be better if we can first show the problem description to the user, and generate the file if the user has some further confirmation action. What do you think?

xuzaixian:
Hum....maybe we can preview the problem within a webview , and generate the file after confirmation action ?

jdneo:
@xuzaixian Exactly what I'm thinking about. And this action can be bound with a new command called `Preview Problem`.

Would you like to have a try to implement this feature first?

xuzaixian:
ok, i will work on it.

jdneo:
@xuzaixian I'm thinking that if we should create a new command or just reuse `Show Problem`

xuzaixian:
I prefer to create a new command `Preview Problem` . 

use 'Show Problem' when they want to code right now
use 'Preview Problem' when they want to read description.
we can provide more choices to user 🤣

jdneo:
@xuzaixian So what should we do when user left-click on the node?

xuzaixian:
@jdneo How about just preview the problem description when user left-click?
We can provide an extra button called `Code Now` or something else ..  in preview window,
The user can click button to code right away. 
Also provide a new menu option in right-click menu called `Code Now`.

This sounds convenient.

jdneo:
@xuzaixian Yes, sounds good

jdneo:
Hi @xuzaixian,

I noticed that a lot of diffs is caused by reformating the `package.json`. Would you mind to revert those formatting changes? This could make it easier to review.

Thanks!

xuzaixian:
sorry for that 😢

src/utils/webviewUtils.ts
```diff
@@ -0,0 +1,54 @@
...
+        <style>
+            #solve {
+                position: fixed;
+                bottom: 10px;
+                right: 10px;
+                border: 0;
+                margin: 10px 0;
+                padding: 2px 14px;
+                color: white;
+                background-color: rgb(14, 99, 156);
```
jdneo: Hard code the color is not recommended because the appearance will not be consistent when the user switches the theme. 

Instead, you can use the VS Code embed vars, for example `background-color: var(--vscode-editor-background);`

If you want to see all the available variables, open a web view, trigger the command: `Developer: Open Webview Developer Tools`, inspect an element, then you can see all the available variables.

image

src/utils/webviewUtils.ts
```diff
@@ -0,0 +1,54 @@
...
+            <meta name="viewport" content="width=device-width, initial-scale=1.0">
+            <title>Preview Problem</title>
+        </head>
+        <style>
+            #solve {
+                position: fixed;
+                bottom: 10px;
+                right: 10px;
+                border: 0;
+                margin: 10px 0;
```
jdneo: Use `rem` instead of `px`: https://stackoverflow.com/a/43131958

src/commands/show.ts
```diff
@@ -135,3 +136,21 @@ async function resolveRelativePath(value: string, node: IProblem, selectedLangua
             throw new Error(errorMsg);
     }
 }
+
+export async function previewProblem(node: IProblem): Promise<void> {
```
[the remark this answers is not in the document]
xuzaixian: ok

src/explorer/LeetCodeNode.ts
```diff
@@ -48,4 +49,20 @@ export class LeetCodeNode {
     public get parentName(): string {
         return this.parentNodeName;
     }
+
+    public get previewCommand(): Command {
```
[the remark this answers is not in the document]
xuzaixian: I thought this pr is about we can get a webview to preivew the problem when we left-click the node.
So... I'm not sure about what behavior you refered 😕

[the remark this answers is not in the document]
jdneo: Hmm, I was confused before.

So the current behavor is:
- Right click the problem will have two open: `Show`, `Preview`
- Left-click the problem will `Preview`

Right?

If yes, it looks good to me.

src/explorer/LeetCodeNode.ts
```diff
@@ -48,4 +49,20 @@ export class LeetCodeNode {
...
+        return {
+            title: "Preview Problem",
+            command: "leetcode.previewProblem",
+            arguments: [this],
+        };
+    }
+
+    public get selectedCommand(): Command {
```
jdneo: So I guess this method can be removed?

src/explorer/LeetCodeNode.ts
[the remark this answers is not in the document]
xuzaixian: Yes.
The current behavor is exactly what you said.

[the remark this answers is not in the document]
jdneo: Cool...

jdneo:
@xuzaixian You can install a extension called: TS Lint, to help you resolve those lint errors.

https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-typescript-tslint-plugin

xuzaixian:
@jdneo 
There is no error in my environment when i run `vsce package`
But The Travis CI build failed.
Can you take a look ?

jdneo:
@xuzaixian It's because the file system on Linux is case sensitive.

your file name is `leetcodePreviewProvider.ts`, but when you import it, you are using `leetCode`.

src/leetcodePreviewProvider.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public dispose(): void {
+        if (this.panel) {
+            this.panel.dispose();
+        }
+    }
+
+    private async provideHtmlContent(node: IProblem): Promise<string> {
+        return await renderHTML(node);
```
jdneo: For now, it's not necessary to extract separate utility, let's move the logic in `webviewUtils.ts` to here

src/utils/webviewUtils.ts
```diff
@@ -0,0 +1,54 @@
...
+                (function() {
+                    const vscode = acquireVsCodeApi();
+                    let button = document.getElementById('solve');
+                    button.onclick = solveHandler;
+                    function solveHandler() {
+                        vscode.postMessage({
+                            command: 'ShowProblem',
+                        })
+                    }
+                }())
```
jdneo: missing `;`

src/utils/webviewUtils.ts
```diff
@@ -0,0 +1,54 @@
...
+            <button id="solve">Code Now</button>
+            <script>
+                (function() {
+                    const vscode = acquireVsCodeApi();
+                    let button = document.getElementById('solve');
+                    button.onclick = solveHandler;
+                    function solveHandler() {
+                        vscode.postMessage({
+                            command: 'ShowProblem',
+                        })
```
jdneo: missing `;`

src/shared.ts
```diff
@@ -91,3 +91,7 @@ export enum Category {
     Company = "Company",
     Favorite = "Favorite",
 }
+
+export interface IWebViewMessage {
```
jdneo: Move the definition to `leetcodePreviewProvider.ts`

Issue #184, opened 2019-03-04
Add an `All` problem list


taoweicn:
## 🚀 Feature Proposal

Add a new problem list which contains all leetcode problems.

## Motivation

Maybe someone such as me wants to solve problems in order.

yucaowang:
+1

jdneo:
@Vigilans I'm thinking to refactor the explorer before adding new features to it. For example, extract the logic of node reference to a cache called something like `xxxCache` or `nodeManager`...

Issue #96, opened 2019-01-24
New features wishlist


Vigilans:
- [ ] When clicking the problem node, create a webview panel with problem description.
This feature could be implemented with `leetcode-cli show -x` command. 

- [ ] Add/Remove favorite problem from right-click context menu.
Could be implemented by `leetcode-cli star` command.

- [ ] Show most voted solution in a webview panel from right-click context menu.
Could be implemented by `leetcode-cli show <id> --solution` command from [solution.dicuss](https://github.com/skygragon/leetcode-cli-plugins/blob/master/docs/solution.discuss.md) plugin.

- [ ] Open problem page in web browser from right-click context menu.

- [ ] Intellisense, lint, debug with testcase locally.
Relative work: `leetcode-cli test <filename> --local` command from [cpp.run](https://github.com/skygragon/leetcode-cli-plugins/blob/master/docs/cpp.run.md) plugin. 
Intellisense and lint service could be provided by other extensions like `cpptool`. What we should do is to provide some pre-included headers/packages like `bits/stdc++.h` and some `ListNode` definitions, which has been covered by the effort of [cpp.run](https://github.com/skygragon/leetcode-cli-plugins/blob/master/plugins/cpp.run.js) plugin.

jdneo:
@Vigilans Great work! All of these features are very useful from my perspective.

taoweicn:
Consider these?
#59 #69

Vigilans:
@taoweicn most of the `leetcode.com` relative features are implemented based on `leetcode-cli`, I'm not sure whether `show the latest submission` feature is implemented there. 

Since I am participating in MCM contest these days(25-29), I might not be able to quickly response the issues...after finishing the contest, I will look into all these features.

20manas:
Is there any update on the fifth milestone? It'd be quite useful to have it.

PR #188, opened 2019-03-06
Star Problems Extension && All Problems Category

Vigilans:
## Introduction

Solved following issues:
* Star Problems: #96 
* All Problems: #184 

### Feature
* `LeetCodeNode` now exposes its inner `IProblem` data through `nodeData` property.
* `TreeDataProvider` now preserves a `allProblems` field which holds the reference of all problems.
* `TreeDataProvider` now provides ability to update one specific problem data with the help of `allProblems` field.
* The starred problem will be appended with "♥" in the tree item name.
* The hovering tooltip is enhanced to work for `All Problems` Category and `Favorite` Category.

### Flaw
* After star/unstar problems, the tree view return to uncollapsed state. It is because we do not store collapse state in the `TreeDataProvider`.

## Demonstration
image
image
image


jdneo:
@Vigilans Thank you for your contribution, will take a look later~

package.json
```diff
@@ -94,6 +94,11 @@
                 "title": "Show Problem",
                 "category": "LeetCode"
             },
+            {
+                "command": "leetcode.starProblem",
```
jdneo: leetcode.toggleFavorite?

src/explorer/LeetCodeTreeDataProvider.ts
```diff
@@ -31,6 +33,14 @@ export class LeetCodeTreeDataProvider implements vscode.TreeDataProvider<LeetCod
         this.onDidChangeTreeDataEvent.fire();
     }
 
+    public async updateProblem(problem: IProblem): Promise<void> {
+        if (this.allProblems.has(problem.id)) {
+            this.updateTreeDataByProblem(problem); // only modify the content of tree data, problem is not updated.
+            Object.assign(this.allProblems.get(problem.id), problem); // update problem, where reference is preserved.
+            this.onDidChangeTreeDataEvent.fire();
```
jdneo: I guess that the reason that explorer returns to uncollapsed state is that, we do not pass the node element into `this.onDidChangeTreeDataEvent.fire()` (This API can accept a node element. Then all the data are cleared before the explorer rerendering.

src/explorer/LeetCodeTreeDataProvider.ts
```diff
@@ -46,7 +56,7 @@ export class LeetCodeTreeDataProvider implements vscode.TreeDataProvider<LeetCod
 
         const idPrefix: number = Date.now();
         return {
-            label: element.isProblem ? `[${element.id}] ${element.name}` : element.name,
+            label: element.isProblem ? `[${element.id}] ${element.name} ${element.isFavorite ? "♥" : ""}` : element.name,
```
jdneo: Curious about the `♥` appearance on different OS platforms, haha.

src/explorer/LeetCodeTreeDataProvider.ts
```diff
@@ -13,6 +13,8 @@ import { LeetCodeNode } from "./LeetCodeNode";
 
 export class LeetCodeTreeDataProvider implements vscode.TreeDataProvider<LeetCodeNode> {
 
+    private allProblems: Map<string, IProblem>; // store reference of all problems.
```
jdneo: Any reason we separate `allProblems` with `treeData`?

src/explorer/LeetCodeTreeDataProvider.ts
```diff
@@ -197,6 +230,18 @@ export class LeetCodeTreeDataProvider implements vscode.TreeDataProvider<LeetCod
         ].join(os.EOL);
     }
 
+    private updateTreeDataByProblem(problem: IProblem): void {
+        const origin: IProblem | undefined = this.allProblems.get(problem.id);
+        if (origin && origin.isFavorite !== problem.isFavorite) {
+            const problemIndex: number = this.treeData.Favorite.findIndex((p: LeetCodeNode) => Number(p.id) >= Number(problem.id));
```
jdneo: can we directly compare `p.id` with `problem.id`?

src/explorer/LeetCodeTreeDataProvider.ts
```diff
@@ -197,6 +230,18 @@ export class LeetCodeTreeDataProvider implements vscode.TreeDataProvider<LeetCod
         ].join(os.EOL);
     }
 
+    private updateTreeDataByProblem(problem: IProblem): void {
```
jdneo: The method name her is confusing. Cuz we only handle `isFavorite` here, but the method name looks like we can update all the fields.

src/commands/star.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) jdneo. All rights reserved.
+// Licensed under the MIT license.
+
+import { LeetCodeNode } from "../explorer/LeetCodeNode";
+import { LeetCodeTreeDataProvider } from "../explorer/LeetCodeTreeDataProvider";
+import { leetCodeExecutor } from "../leetCodeExecutor";
+import { IProblem } from "../shared";
+import { DialogType, promptForOpenOutputChannel } from "../utils/uiUtils";
+
+export async function starProblem(provider: LeetCodeTreeDataProvider, node: LeetCodeNode): Promise<void> {
```
jdneo: toggleFavorite?

src/leetCodeExecutor.ts
```diff
@@ -87,6 +87,16 @@ class LeetCodeExecutor {
         return filePath;
     }
 
+    public async starProblem(node: IProblem, markStarred: boolean): Promise<boolean> {
```
jdneo: `toggleFavorite(...)`?

package.json
Vigilans: A much better choice!

src/explorer/LeetCodeTreeDataProvider.ts
Vigilans: The problems are originally in number-ascending order, thus the comparation in index searching should also be done this way. 
e.g. the favorite problems' ids are ["1", "4", "16", "23"], I want to delete the problem with id "23" and id is compared directly. Since "4" >= "23" === true, the search process stops at "4", thus the deletion is incorrect.

src/explorer/LeetCodeTreeDataProvider.ts
Vigilans: I thought we might handle the case of updating other `treeData` fields the at a later time. Seems a overdesign here?
Should I leave a comment announcing `modification of other categories is not needed yet` here, or just change the method name?

src/explorer/LeetCodeTreeDataProvider.ts
Vigilans: After some investigations I seems to get the hang of this API. I think maybe we should keep another reference container here:
```typescript
allTreeNodes: Map<string, LeetcodeNode[]> // id -> relative nodes
```
Then, in `updateProblem` function, we can fire events to request all relative nodes to be updated.

src/explorer/LeetCodeTreeDataProvider.ts
Vigilans: * `allProblems` is an id-IProblem dictionary primarily for fast lookup, whose structure is not compatible with `treeData`;
* `allProblems` serves as a buffer which holds all the problems' ownership, where elements in `treeData` is just references to the problems;
* with regard to `this.onDidChangeTreeDataEvent.fire(problem)` review below, we may need another lookup dictionary for fast retrieving relative nodes. `allProblems` seems more similar to it rather than `treeData`.

src/explorer/LeetCodeTreeDataProvider.ts
jdneo: @Vigilans Cool. Noticed that now the same problem in the different category has different id. So if we refresh the `Two Sum` in `All Problem`, will the `Two Sum` in `Difficulty` be refreshed?

That means, I'm thinking that if we should make the same problem has the same id in the whole explorer. Not sure the real behavior of it.

src/explorer/LeetCodeTreeDataProvider.ts
jdneo: IMO, the main problem here is that we store the Favorite problems as an array instead of a Map(id -> Problem).

So you may observe that we have to use `findIndex` and `splice` to add/remove items, which I think is hard to read.

Can we change it to a Map?

src/explorer/LeetCodeTreeDataProvider.ts
Vigilans: Make sure now other fields will be updated here in the near PRs, e.g. when a problem is accepted, there will be some code here to update other fields.

Vigilans:
In order to hide sovled problems in `All Problems` Category, `Category.All` is added as a field to `treeData`.

Now, `allProblems` holds the ownership of every problems, whereas `treeData[Category.All]` may hide problems that is solved.

src/explorer/LeetCodeTreeDataProvider.ts
Vigilans: After further investigations, to accomplish this feature, there are some important changes which I think should be discussed in next PR. In this PR, we may temporarily accept the current behavior?

src/explorer/LeetCodeTreeDataProvider.ts
Vigilans: Now `allProblems` is decoupled with `treeData`, and will be grouped with `allTreeNodes` as an object pool in next PR.

src/explorer/LeetCodeTreeDataProvider.ts
Vigilans: Maybe it can be discussed in subsequent PRs, when updating `treeData[Category.Company/Tag]` field is necessary(delete problem from treeview after it is solved)?

src/explorer/LeetCodeTreeDataProvider.ts
```diff
@@ -42,71 +53,75 @@ export class LeetCodeTreeDataProvider implements vscode.TreeDataProvider<LeetCod
                     title: "Sign in to LeetCode",
                 },
             };
+        } else if (!element.isProblem) { // category
+            return {
+                label: element.name,
+                tooltip: this.getSubCategoryTooltip(element),
+                id: `LeetCode.Category::${element.parentId}.${element.id}`,
```
Vigilans: Now that we have `element.parentId`, `Data.Now()` prefix seems no longer necessary.

Btw, the TreeItem Generation code here is separated here for better readability, as well as convenience for next PR(store generated TreeItem somewhere in TreeProvider or LeetCodeNode)

src/explorer/LeetCodeTreeDataProvider.ts
jdneo: Yes, can be solved in another PR.

jdneo:
I'll take a deep look at the weekend.

src/explorer/LeetCodeTreeDataProvider.ts
```diff
@@ -13,26 +13,37 @@ import { LeetCodeNode } from "./LeetCodeNode";
...
-        Favorite: IProblem[],
+        [Category.All]: IProblem[],
+        [Category.Difficulty]: Map<string, IProblem[]>,
+        [Category.Tag]: Map<string, IProblem[]>,
+        [Category.Company]: Map<string, IProblem[]>,
+        [Category.Favorite]: IProblem[],
     };
 
-    private onDidChangeTreeDataEvent: vscode.EventEmitter<any> = new vscode.EventEmitter<any>();
+    private onDidChangeTreeDataEvent: vscode.EventEmitter<LeetCodeNode> = new vscode.EventEmitter<LeetCodeNode>();
```
jdneo: We can also support refresh the whole explorer, so the parameter type can be written as `LeetCodeNode | null | undefined`

src/explorer/LeetCodeTreeDataProvider.ts
jdneo: So you will send out another PR to remove `element.parentId` & `Data.Now()`, right?

src/explorer/LeetCodeTreeDataProvider.ts
jdneo: So `updateProblem()` is not used in this PR, right?

src/explorer/LeetCodeTreeDataProvider.ts
Vigilans: The event fire's parameter is `fire(data: T?)` where T=LeetCodeNode, so no need to append null and undefined here

src/explorer/LeetCodeTreeDataProvider.ts
Vigilans: Data.Now() has already been removed here, which is taken place by element.parentId

Vigilans:
There are so many interesting and important things about treeview...which will make our explorer far better than current form.

### TreeItem.id & Collapse State
* [Selection and expansion state is lost after tree node label change](https://github.com/Microsoft/vscode/issues/40018)
* [Expose the collapsible state of the tree item](https://github.com/Microsoft/vscode/issues/30535)

### TreeItem.description
* [API to show description for tree item labels](https://github.com/Microsoft/vscode/issues/62795)
* [Test Tree API additions](https://github.com/Microsoft/vscode/issues/64205)
```
/**
 * A human readable string which is rendered less prominent.
 * When `true`, it is derived from [resourceUri](#TreeItem.resourceUri) and when `falsy`, it is not shown.
 */
description?: string | boolean;
```
This field is officially public since vscode engine 1.30, and `gitlens` has already adopted it:
image
May be we can show some message including difficulty and pass rate to solve #192.

### TreeItem.highlight
* [TreeItem should support highlights](https://github.com/Microsoft/vscode/issues/61313)
* [Introduce tree item highlights](https://github.com/Microsoft/vscode/pull/61482)
This API is now in proposed API. Maybe after enabling the proposed API in package.json, we can use it to render the problems by their difficutly.

### Reference Repo
[vscode-sample](https://github.com/Microsoft/vscode-extension-samples/blob/master/tree-view-sample/src/ftpExplorer.ts)
[GitLens](https://github.com/eamodio/vscode-gitlens/)

I think this PR could be temporarily suspended for further discussion, or merge it for now and then progressively refactoring the TreeDataProvider.

jdneo:
@Vigilans looks like we have a lot of works to do with the explorer. Yes, my personal suggestion is to suspend the PR just for now and implement the improvement one by one.

You can separately create issues for each of them, and we can discuss the implementation in each issue before sending out the PR.

Allianzcortex:
If we can add more search options(e.g. from easy->difficult) when displaying the problems in one category ,it can be awesome . The order of problems up to now is based on number order.

jdneo:
@Vigilans 

Seems this PR is out of date. Shall we close this and open another PR about `toggle the favorite problem`? (Show all problems have been resolved)

Vigilans:
@jdneo Yes, and there are some more improvements that could be done on `toggleFavorite`, like using a button to toggle.

PR #193, opened 2019-03-08
Top voted solution webview

Vigilans:
## Introduction

Solves #96 
This PR is independent with #188.

## Details
* indented code block uses solution language to highlight.
* uses built-in markdown extension to provide style files.

## Demonstration
image

image

## References

* [electron-browser code template](https://github.com/Microsoft/vscode/blob/eb12756d0c3cb8c20a7748986713c8a40bc4afd5/src/vs/workbench/contrib/extensions/electron-browser/extensionEditor.ts#L59)
* [built-in markdown extension](https://github.com/Microsoft/vscode/tree/038e8d39046fd93923785f9bcfe9d8073fe6a4f1/extensions/markdown-language-features/src)


jdneo:
@Vigilans Sorry for the conflict made by merging another PR, would you mind to resolve it?

Vigilans:
@jdneo Conflict has been resolved.

jdneo:
Cool, I'll take a look tonight

src/leetCodeExecutor.ts
```diff
@@ -87,6 +89,11 @@ class LeetCodeExecutor {
         return filePath;
     }
 
+    public async showSolution(node: IProblem, language: string): Promise<string> {
```
jdneo: rename the param `node` to `problemNode` or `problem` to avoid get confusing with the command `node`

src/shared.ts
```diff
@@ -85,6 +85,15 @@ export const defaultProblem: IProblem = {
     tags: [] as string[],
 };
 
+export class Solution {
```
jdneo: Seems only `src/leetCodeSolutionProvider.ts` referenced the Solution, put it into the `leetCodeSolutionProvider`

src/leetCodeSolutionProvider.ts
```diff
@@ -0,0 +1,124 @@
...
+        return stylePaths.map((p: string) => vscode.Uri.file(path.join(this.markdownPath, p)).with({ scheme: "vscode-resource" }));
+    }
+
+    private getWebViewContent(solution: Solution): string {
+        const styles: string = this.getMarkdownStyles()
+            .map((style: vscode.Uri) => `<link rel="stylesheet" type="text/css" href="${style.toString()}">`)
+            .join("\n");
+        const { title, url, lang, author, votes } = solution;
+        const head: string = this.markdown.render(`# [${title}](${url})`);
+        const auth: string = `[${author}](https://leetcode.com/${author}/)`;
```
jdneo: So the solutions are always from the English version of LeetCode right?

Vigilans: After test, the `solution.discuss` indeed fetch solutions only from leetcode.com. After switching to our own library, we may be able to freely choose endpoint.

src/leetCodeSolutionProvider.ts
```diff
@@ -0,0 +1,134 @@
...
+    private solution: Solution;
+
+    public initialize(context: ExtensionContext): void {
+        this.context = context;
+        this.markdown = new MarkdownIt({
+            linkify: true,
+            typographer: true,
+            highlight: this.codeHighlighter.bind(this),
+        });
+        this.markdownPath = path.join(process.env.VSCODE_CWD as string, "resources", "app", "extensions", "markdown-language-features");
```
jdneo: Use `vscode.env.appRoot` instead of the environment variable since it will not always be available

src/leetCodeSolutionProvider.ts
```diff
@@ -0,0 +1,134 @@
...
+        if (hljs.getLanguage(lang)) {
+            try {
+                return hljs.highlight(lang, code, true).value;
+            } catch (error) { /* do not highlight */ }
+        }
+        return ""; // use external default escaping
+    }
+
+    private getMarkdownStyles(): vscode.Uri[] {
+        const stylePaths: string[] = require(path.join(this.markdownPath, "package.json"))["contributes"]["markdown.previewStyles"];
```
jdneo: Emm.. Got `Cannot find module '/Applications/Visual Studio Code.app/Contents/Resources/app/resources/app/extensions/markdown-language-features/package.json'` here.

While I can confirm that this path does exist, but have no idea why it could not be found. Do you know why?

I'm thinking that at least we should add a try-catch block here and return `[]` in the catch in case that we cannot get the style paths

src/leetCodeSolutionProvider.ts
Vigilans: Yes, you are right. The VSCODE_CMD does not exist when I debug on my Linux machine.

src/leetCodeSolutionProvider.ts
Vigilans: No idea what happened... After I switch to `vscode.env.appRoot`, the code both works on my Win and Linux machine. 

Could you have one more try one the new commit? If the problem still exists, I'll go somewhere to find a Mac to look into the problem.

src/leetCodeExecutor.ts
```diff
@@ -48,10 +48,11 @@ class LeetCodeExecutor {
             }
             return false;
         }
-        try { // Check company plugin
-            await this.executeCommandEx("node", [await this.getLeetCodeBinaryPath(), "plugin", "-e", "company"]);
-        } catch (error) { // Download company plugin and activate
-            await this.executeCommandEx("node", [await this.getLeetCodeBinaryPath(), "plugin", "-i", "company"]);
+        for (const plugin of ["company", "solution.discuss", "leetcode.cn"]) { // Make sure plugin exists
```
jdneo: I remember `-i` will automatically enable it, is that right?

src/leetCodeSolutionProvider.ts
Vigilans: image
I still got the style work on a Mac...Yet it may be worth noting that there is `Users/<username>` before `Application`.
If it means that your code is installed in root applications, then does it have anything to do with the file access authorization?

src/leetCodeExecutor.ts
Vigilans: The `getLeetCodeEndpoint` returns the configuration from vscode. No matter `leetcode.cn` is enabled or not, the vscode configuration will rectify it.

src/leetCodeSolutionProvider.ts
jdneo: @Vigilans My fault, I forget to remove the `resources/app`, sorry for the confusion, haha

src/leetCodeExecutor.ts
jdneo: Yes, it's because of the cache.

Still cannot see the purpose to change the logic here, the original implementation looks fine to me.

jdneo:
@Vigilans Thank you for keep working on this PR, it's very near to get merged.

src/leetCodeExecutor.ts
Vigilans: For unknown reason, when  I debug the extension on my Linux machine, it pops out `PLUGIN NOT FOUND` on the execution of `"node", [await this.getLeetCodeBinaryPath(), "plugin", "-e", "leetcode.cn"]`, rather than download it and install.  But after I install the plugin, everything works fine again.

src/leetCodeSolutionProvider.ts
```diff
@@ -0,0 +1,140 @@
...
+        }
+        return ""; // use external default escaping
+    }
+
+    private getMarkdownStyles(): vscode.Uri[] {
+        try {
+            const stylePaths: string[] = require(path.join(this.markdownPath, "package.json"))["contributes"]["markdown.previewStyles"];
+            return stylePaths.map((p: string) => vscode.Uri.file(path.join(this.markdownPath, p)).with({ scheme: "vscode-resource" }));
+        } catch (error) {
+            promptForOpenOutputChannel("Fail to load built-in markdown style file.", DialogType.error);
```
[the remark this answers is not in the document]
Vigilans: How to print to output channel without using `executeCommand` series?

src/leetCodeExecutor.ts
jdneo: Is this reproducible on your side?

Vigilans: After installed `leetcode.cn.js`, I tried to delete it, then debug the extension, but the bug can no longer be produced anymore. I guess it can only be encountered at the very first time of using `leetcode-cli`.

src/leetCodeSolutionProvider.ts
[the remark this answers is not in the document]
jdneo: Change `promptForOpenOutputChannel("Fail to load built-in markdown style file.", DialogType.error);` 

to 

`leetCodeChannel.appendLine("[Error] Fail to load built-in markdown style file.");`

src/leetCodeExecutor.ts
jdneo: I suggest we leave the code as it's original implementation. If this is a problem and users open issues for it, we can address it in another PR?

jdneo: This could also be an error accidentally thrown by executing `leetcode plugin -i xx`, who knows.

Vigilans: Reasonable suggestion, updated

PR #197, opened 2019-03-09
vscode versions to v1.31.0, add collapes all.

ZainChen:
[collapes all](https://images.cnblogs.com/cnblogs_com/chenzhiyin/1399691/o_Leetcode.png)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,8 +55,7 @@
         "acorn": {
             "version": "2.7.0",
             "resolved": "https://registry.npmjs.org/acorn/-/acorn-2.7.0.tgz",
-            "integrity": "sha1-q259nYhqrKiwhbwzEreaGYQz8Oc=",
-            "optional": true
+            "integrity": "sha1-q259nYhqrKiwhbwzEreaGYQz8Oc="
         },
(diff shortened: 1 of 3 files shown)


ZainChen:
[collapes all](https://github.com/jdneo/vscode-leetcode/pull/197)

jdneo:
@ZainChen Thank you for your contribution.

The feature `collapes all` looks good to me, just on concern about it.

Could you make sure that `1.31.0` is the lowest version to support this API? Cuz by upgrading the engine version, all the users who are using VS Code lower than `1.31.0` could not use the extension anymore.

ZainChen:
The minimum version that supports this API is 1.30.1. Although the official 1.29.1 version pointed out the modifications, the interface appeared in version 1.30.1.
https://code.visualstudio.com/updates/v1_29
https://code.visualstudio.com/updates/v1_30

ZainChen:
The contribution version is changed to 1.30.1, you can prompt everyone to use the new version of vscode, everyone will like this feature.^_^

ZainChen:
Thank you too, very happy to contribute to such a good expansion.

PR #202, opened 2019-03-12
[bugfix] Correct the  content shown in status-bar

### LINKED ISSUE #112: Wrong user name shown in the status bar
When activate the extension, sometimes, in the status bar, it will show: `LeetCode: Host`, which should be `LeetCode: <user name>`

jdneo:
Resolve #112

### DIFF
--- a/src/leetCodeManager.ts
+++ b/src/leetCodeManager.ts
@@ -119,13 +119,10 @@ class LeetCodeManager extends EventEmitter {
     }
 
     private tryParseUserName(output: string): string {
-        const lines: string[] = output.trim().split(/\r?\n/);
-        if (lines.length === 3) {
-            const reg: RegExp = /^\s*.{1}\s*(.+)\s*https:\/\/leetcode/;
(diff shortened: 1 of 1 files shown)


src/leetCodeManager.ts
```diff
@@ -117,6 +117,19 @@ class LeetCodeManager extends EventEmitter {
     public getUser(): string | undefined {
         return this.currentUser;
     }
+
+    private tryParseUserName(output: string): string {
+        const lines: string[] = output.trim().split(/\r?\n/);
+        if (lines.length === 3) {
+            const reg: RegExp = /^\s*.{1}\s*(.+)\s*https:\/\/leetcode/;
```
Vigilans: The regex executing on 
```
'    ×      vigilans             https://leetcode-cn.com'
```
will return
```
'vigilans             '
```
Append `?` to `.+` to prevent greedy match could solve the problem.

Besides, is there any difference between `.{1}` and `.`?

jdneo: Updated

PR #204, opened 2019-03-13
Prepare for 0.13.0

jdneo:


README.md
```diff
@@ -70,14 +70,23 @@
   <img src="https://raw.githubusercontent.com/jdneo/vscode-leetcode/master/docs/imgs/pick_problem.png" alt="Pick a Problem" />
 </p>
 
-- Right click the problem in the `LeetCode Explorer` and select `Show Problem` will generate a new file with the problem description for you.
+- Right click the problem in the `LeetCode Explorer` and select `Preview Problem` to see the problem description. Or select `Show Problem` to directly open the file with the problem description.
```
Vigilans: Left click the problem node would also trigger preview window, isn't it necessary to notify this?

jdneo: Good catch, I'll add this.

jdneo:
yes, we can improve the rendering logic of preview and result in the later version, it's not a big problem.

PR #216, opened 2019-03-17
[bugfix] Clicking 'Code Now' will open multiple files

### LINKED ISSUE #217: Clicking 'Code Now' will open multiple files
### Bug repro step:

1. Click 2-3 problem node to preview the description one by one
2. Click Code Now
3. It will generate all the files for the problems that have been previewed before

### Root cause:
The listeners are registered every time executing preview (). They should be registered only once when this.panel has not been created.

jdneo:
Fix #217

Bug repro step:

1. Click 2-3 problem node to preview the description one by one
2. Click `Code Now`
3. It will generate all the files for the problems that have been previewed before

Root cause:
The listeners are registered every time executing `preview ()`. They should be registered only once when `this.panel` has not been created.

### DIFF
--- a/src/leetCodePreviewProvider.ts
+++ b/src/leetCodePreviewProvider.ts
@@ -4,6 +4,7 @@ import { IProblem } from "./shared";
 class LeetCodePreviewProvider implements Disposable {
 
     private context: ExtensionContext;
+    private node: IProblem;
     private panel: WebviewPanel | undefined;
 
     public initialize(context: ExtensionContext): void {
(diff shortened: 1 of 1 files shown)


Vigilans:
I tested the bug on my side: 
Click problem [3], [70], [169], and the behavior appears as expected. Besides, after testing the behavior multiple times, one of the problem may fail to be shown for `JSON parse error on input`. After this PR, this problem may disappear.

Other flaws of the preview command:
* the extension will still fetch problem description after clicking the active problem node
* did not make use of local cache

jdneo:
Merged for now. We can improve both the `preview` and the `result` webview in `0.13.2` using the markdown solution, which I think can improve the experience significantly. Meanwhile, more bugs and some refactoring work can be addressed in `0.13.2`.

Issue #205, opened 2019-03-13
Solution webview doesn't highlight python3 language


Vigilans:
## 🐛 Bug Report

The `highlight.js` seems unable to recognize python3, and will not highlight the code fences.

## To Reproduce

Execute `Show Top Voted Solution` on any problem with language python3, then we'll get
![image](https://user-images.githubusercontent.com/20227484/54288145-ea28a080-45e1-11e9-95bd-9d405b912eb9.png)

## Expected behavior

Alter `Solution.lang: python3` to `python` when passed to `highlight.js`.

jdneo:
@Vigilans I think this should be fixed in #224, right?

Vigilans:
@jdneo Yes, I have tested it.

PR #224, opened 2019-03-19
Refactor solution webview to reuse markdown engine

Vigilans:
## Introduction

* Extract markdown rendering logic outside of solution webview as `MarkdownEngine` class.
* Solved #205, solving https://github.com/jdneo/vscode-leetcode/pull/216#issuecomment-473664140

## Details

* Add link validator logic, making it support `file://` protocal.
* Enhanced indented code block highlight logic.
* Correction of multiple unsupported alias of language(e.g. python3)
* Completion of image url in solution markdown(e.g. /asset/... -> https://discuss.leetcode.com/asset/..., refer to [this solution](https://leetcode.com/problems/bulb-switcher-ii/discuss/107277))


src/leetCodeSolutionProvider.ts
```diff
@@ -1,49 +1,31 @@
...
-        this.markdown.renderer.rules["code_block"] = (tokens, idx, options, _, self) => {
-            const highlight: string = options.highlight(tokens[idx].content, undefined);
-            return [
-                `<pre><code ${self.renderAttrs(tokens[idx])} >`,
-                highlight || this.markdown.utils.escapeHtml(tokens[idx].content),
-                "</code></pre>",
-            ].join("\n");
-        };
+        this.markdown = new MarkdownEngine();
+        this.addMdDefaultHighlight(); // use solution language if highting block lang is undefined
```
jdneo: `highting ` typo?

src/webview/markdownEngine.ts
```diff
@@ -0,0 +1,99 @@
...
+            return [];
+        }
+    }
+
+    public get options(): MarkdownIt.Options {
+        return (this.engine as any).options;
+    }
+
+    public getStylesHTML(): string {
+        return this.styles.map((style: vscode.Uri) => `<link rel="stylesheet" type="text/css" href="${style.toString()}">`).join("\n");
```
jdneo: nit: `\n` -> `os.EOL`

src/leetCodeSolutionProvider.ts
```diff
@@ -63,6 +45,27 @@ class LeetCodeSolutionProvider implements Disposable {
...
+    private addMdDefaultHighlight(): void {
+        // The @types typedef of `highlight` is wrong, which should return a string.
+        // tslint:disable-next-line:typedef
+        const highlight = this.markdown.options.highlight as (code: string, lang?: string) => string;
+        this.markdown.options.highlight = (code: string, lang?: string): string => {
+            return highlight(code, lang || this.solution.lang);
+        };
+    }
+
+    private addMdImageUrlCompletion(): void {
```
jdneo: What is this method for? Looks like this method could be moved into the  MarkDownEngine.

src/webview/markdownEngine.ts
```diff
@@ -0,0 +1,99 @@
+import * as hljs from "highlight.js";
```
jdneo: file name change to `MarkdownEngine` since it's the only class exported in this file.

Also, remember to correct the import statement to avoid compilation error

src/webview/markdownEngine.ts
```diff
@@ -0,0 +1,99 @@
...
+import * as MarkdownIt from "markdown-it";
+import * as path from "path";
+import * as vscode from "vscode";
+import { leetCodeChannel } from "../leetCodeChannel";
+
+export class MarkdownEngine {
+
+    public readonly engine: MarkdownIt;
```
jdneo: private

src/webview/markdownEngine.ts
```diff
@@ -0,0 +1,99 @@
...
+    public readonly extRoot: string; // root path of vscode built-in markdown extension
```
jdneo: private

src/leetCodeSolutionProvider.ts
```diff
@@ -63,6 +45,27 @@ class LeetCodeSolutionProvider implements Disposable {
         }
     }
 
+    private addMdDefaultHighlight(): void {
+        // The @types typedef of `highlight` is wrong, which should return a string.
+        // tslint:disable-next-line:typedef
+        const highlight = this.markdown.options.highlight as (code: string, lang?: string) => string;
+        this.markdown.options.highlight = (code: string, lang?: string): string => {
+            return highlight(code, lang || this.solution.lang);
```
jdneo: Under what circumstances will the param `lang` be undefined?

src/webview/markdownEngine.ts
```diff
@@ -0,0 +1,99 @@
...
+                "</code></pre>",
+            ].join("\n");
+        };
+    }
+
+    private addLinkValidator(md: MarkdownIt): void {
+        const validateLink: (link: string) => boolean = md.validateLink;
+        md.validateLink = (link: string): boolean => {
+            // support file:// protocal link
+            return validateLink(link) || link.startsWith("file:");
```
jdneo: Could you explain more why we need the validator here?

src/webview/markdownEngine.ts
```diff
@@ -0,0 +1,99 @@
...
+            if (tokens.some((token: any) => token.type === "fence")) {
+                return origin(tokens, idx, options, env, self);
+            }
+            // otherwise, highlight with undefined lang, which could be handled in outside logic.
+            const highlighted: string = options.highlight(tokens[idx].content, undefined);
+            return [
+                `<pre><code ${self.renderAttrs(tokens[idx])} >`,
+                highlighted || md.utils.escapeHtml(tokens[idx].content),
+                "</code></pre>",
+            ].join("\n");
```
jdneo: os.EOL

src/webview/markdownEngine.ts
```diff
@@ -0,0 +1,99 @@
...
+            },
+        });
+
+        this.addCodeBlockHighlight(md);
+        this.addLinkValidator(md);
+        return md;
+    }
+
+    private addCodeBlockHighlight(md: MarkdownIt): void {
+        const origin: MarkdownIt.TokenRender = md.renderer.rules["code_block"];
```
jdneo: origin -> originRender

src/webview/markdownEngine.ts
```diff
@@ -0,0 +1,99 @@
...
+    public render(md: string, env?: any): string {
+        return this.engine.render(md, env);
+    }
+
+    private initEngine(): MarkdownIt {
+        const md: MarkdownIt = new MarkdownIt({
+            linkify: true,
+            typographer: true,
+            highlight: (code: string, lang?: string): string => {
+                if (lang && ["tsx", "typescriptreact"].indexOf(lang.toLocaleLowerCase()) >= 0) {
```
jdneo: I think using `switch-case` block will be better than the multiple `if` blocks

src/leetCodeSolutionProvider.ts
jdneo: The two methods `addMdDefaultHighlight()` and `addMdImageUrlCompletion()` is somehow confusing here. The purpose of this PR is to extract and decouple the markdown related code with the webviews, but these two methods are still making them coupling together, which is not a good design.

Looks like the method `addMdImageUrlCompletion()` can be moved into the markdown engine since it does not rely on any member in the solution provider.

For `addMdDefaultHighlight()` we rely on `this.solution.lang` so I asked a question below to see that if it is possible to remove the dependency here.

src/leetCodeSolutionProvider.ts
Vigilans: There are two ways to define a code block:
1. Fence (with lang specification)
> \```lang
> code here
> \```

2. Code block (indented, no lang specification)
> emmm 
> 
>  code here
>
> emmmm

Which exists in the discussions equally frequent.

The `code_block` does not have a language specification, so it will not be highlighted in the markdown-it rule. 

I decided to use the solution language tag to highlight it, so I alter the `code_block` rule, just as what [`MarkdownEngine.addCodeBlockHighlight`](https://github.com/jdneo/vscode-leetcode/pull/224/files#diff-f7ac20b7043db95594bdb6d86f3ef02bR77) do:
1. Since there is no language information in `code_block` rendering rule, so I pass `undefined` to `lang` parameter.
2. Since I have found in some dicussions that there is mixed usage of `fence` and `code_block`, where the author just intended to put some lang-free message in indented `code_block`, so I decided to not highlight the `code_block` if any `fence` is found.

src/webview/markdownEngine.ts
Vigilans: This is the routine adopted by built-in markdown exntesion's [MarkdownEngine](https://github.com/Microsoft/vscode/blob/master/extensions/markdown-language-features/src/markdownEngine.ts#L252), I think it may enables us to validate local file link, which may be useful in local debugger.

src/webview/markdownEngine.ts
Vigilans: This is the routine adopted by built-in markdown extension's [engine](https://github.com/Microsoft/vscode/blob/master/extensions/markdown-language-features/src/markdownEngine.ts#L292). I agree `switch case` is a better choice, and will adopt it in next commit.

src/webview/markdownEngine.ts
Vigilans: Switch to this routine:
```typescript
private readonly _engine: MarkdownIt;
public get engine() { return this.engine; }
```
?

src/leetCodeSolutionProvider.ts
Vigilans: The host name is context-dependent (markdownEngine.ts#L292):
```typescript
imageSrc[1] = `https://discuss.leetcode.com${imageSrc[1]}`; // only valid in Solution webview
```
If I can manage to find a way to insert host name as a context value when rendering the markdown, maybe it can be moved into the MarkdownEngine.

src/leetCodeSolutionProvider.ts
Vigilans: The reason why I merely export `MarkdownEngine` class definition, rather than a global variable `export const markdownEngine = new MarkdownEngine` is that there may be some customization in each WebView, e.g. `this.solution.lang` and url completion here.

Markdown-it may provide some good mechanism for adding custom rule (maybe [Ruler](https://markdown-it.github.io/markdown-it/#Ruler)), but some custom code in each WebView may not be easily avoided.

Vigilans: The workaround may be store `host` and `defaultLang` as an option in MarkdownEngine or somewhere in `MarkdownIt`, then we can move them out of Solution WebView. Do you prefer this way?

src/webview/markdownEngine.ts
Vigilans: If we adopt the workaround (`host` and `defaultLang`) above, maybe there is no need to expose engine and extPath anymore.

src/webview/markdownEngine.ts
Vigilans: How about this way?
```javascript
lang = lang && ({
    mysql: "sql",
    json5: "json",
    python3: "python",
})[lang.toLowerCase()];
```
"tsx" and "typescriptreact" is ruled out, and "mysql" is taken account.

Besides, a language's name is always English, so no need to consider locale mapping.

Vigilans:
Now the markdown engine is fully decoupled with webviews with the help of `env` parameter.

src/leetCodeSolutionProvider.ts
```diff
@@ -1,49 +1,28 @@
...
-        // tslint:disable-next-line:typedef
-        this.markdown.renderer.rules["code_block"] = (tokens, idx, options, _, self) => {
-            const highlight: string = options.highlight(tokens[idx].content, undefined);
-            return [
-                `<pre><code ${self.renderAttrs(tokens[idx])} >`,
-                highlight || this.markdown.utils.escapeHtml(tokens[idx].content),
-                "</code></pre>",
-            ].join("\n");
-        };
+        this.markdown = new MarkdownEngine();
```
jdneo: nit: `markdown` -> `markdownEngine` or `mdEngine` or `engine`.

src/webview/MarkdownEngine.ts
```diff
@@ -0,0 +1,109 @@
...
+        try {
+            const stylePaths: string[] = require(path.join(this.extRoot, "package.json"))["contributes"]["markdown.previewStyles"];
+            return stylePaths.map((p: string) => vscode.Uri.file(path.join(this.extRoot, p)).with({ scheme: "vscode-resource" }));
+        } catch (error) {
+            leetCodeChannel.appendLine("[Error] Fail to load built-in markdown style file.");
+            return [];
+        }
+    }
+
+    public get options(): MarkdownIt.Options {
```
jdneo: seems unused...

src/webview/markdownEngine.ts
jdneo: Ok... Actually still not quite understand... I think it's a worth investigating topic. It's fine to leave it here in this PR...

src/webview/MarkdownEngine.ts
```diff
@@ -0,0 +1,109 @@
+import * as hljs from "highlight.js";
+import * as MarkdownIt from "markdown-it";
+import * as os from "os";
+import * as path from "path";
+import * as vscode from "vscode";
+import { leetCodeChannel } from "../leetCodeChannel";
+
+export class MarkdownEngine {
+
+    public readonly engine: MarkdownIt;
```
jdneo: Can be private now...

src/webview/markdownEngine.ts
Vigilans: Document from markdown-it:

> MarkdownIt#validateLink(url)Boolean

Link validation function. CommonMark allows too much in links. By default we disable `javascript:`, `vbscript:`, `file:` schemas, and almost all `data:...` schemas except some embedded image types.

You can change this behaviour:
```javascript
var md = require('markdown-it')();
// enable everything
md.validateLink = function () { return true; }
```

Since we will deal with local files in WebView, chances are that it will be helpful to enable `file:` link.

src/leetCodeSolutionProvider.ts
Vigilans: Changed to mdEngine

src/webview/MarkdownEngine.ts
Vigilans: Removed

src/webview/MarkdownEngine.ts
Vigilans: I forgot to change it when resolving the conversation...

src/webview/markdownEngine.ts
Vigilans: I looked into the source code, and found that it was made to prevent XSS attack. Here are some examples:
* When parsing image and link:
```javascript
    if (res.ok) {
      href = state.md.normalizeLink(res.str);
      if (state.md.validateLink(href)) {
        pos = res.pos;
      } else {
        href = '';
      }
    }
```
If `validateLink` does not pass, then `href` is screened out.

* The same goes on for reference-link or auto link:
```js
  href = state.md.normalizeLink(res.str);
  if (!state.md.validateLink(href)) { return false; }
```

So, it is indeed necessary to loose the restriction to allow `file://` protocol.

PR #228, opened 2019-03-21
Centralize the webview providers into the webview folder

jdneo:


jdneo:
@Vigilans Cool.. thank you!

PR #232, opened 2019-03-23
Apply markdown engine to result provider

Vigilans:
## Introduction

Solving https://github.com/jdneo/vscode-leetcode/pull/216#issuecomment-473664140

## Demonstration
image

image

image

image

### DIFF
--- a/src/webview/leetCodeResultProvider.ts
+++ b/src/webview/leetCodeResultProvider.ts
@@ -3,7 +3,6 @@
 
 import { Disposable, ExtensionContext, ViewColumn, WebviewPanel, window } from "vscode";
 import { leetCodeChannel } from "../leetCodeChannel";
-import { IProblem } from "../shared";
 import { MarkdownEngine } from "./MarkdownEngine";
 
 class LeetCodeResultProvider implements Disposable {
@@ -49,7 +48,7 @@ class LeetCodeResultProvider implements Disposable {
                     [result.status, raw] = raw.split(/  . (.+)([^]+)/).slice(1);
                     [result.passed, raw] = raw.split(/\n  . (.+)([^]+)/).slice(1);
                     [result.runtime, raw] = raw.split(/\n  . (.+)([^]+)/).slice(1);
-                    [result.memory, raw] = raw.split(/\n  . (.+)/).slice(1);
+                    [result.memory] = raw.split(/\n  . (.+)/).slice(1);
                     return result;
                 }
                 case "×": {
@@ -59,7 +58,7 @@ class LeetCodeResultProvider implements Disposable {
                     [result.testcase, raw] = raw.split(/\n  . testcase: '(.+)'([^]+)/).slice(1);
                     [result.answer, raw] = raw.split(/\n  . answer: (.+)([^]+)/).slice(1);
                     [result.expected, raw] = raw.split(/\n  . expected_answer: (.+)([^]+)/).slice(1);
-                    [result.stdout, raw] = raw.split(/\n  . stdout: ([^]+?)\n$/).slice(1);
+                    [result.stdout] = raw.split(/\n  . stdout: ([^]+?)\n$/).slice(1);
                     result.testcase = result.testcase.replace("\\n", "\n");
                     return result;
                 }
--- a/src/webview/leetCodeResultProvider.ts
+++ b/src/webview/leetCodeResultProvider.ts
@@ -2,7 +2,6 @@
 // Licensed under the MIT license.
 
 import { Disposable, ExtensionContext, ViewColumn, WebviewPanel, window } from "vscode";
-import { leetCodeChannel } from "../leetCodeChannel";
 import { markdownEngine } from "./markdownEngine";
 
 class LeetCodeResultProvider implements Disposable {
@@ -27,7 +26,7 @@ class LeetCodeResultProvider implements Disposable {
             }, null, this.context.subscriptions);
         }
 
-        const result: Result = this.parseResult(resultString);
+        const result: IResult = this.parseResult(resultString);
         this.panel.webview.html = this.getWebViewContent(result);
         this.panel.reveal(ViewColumn.Two);
     }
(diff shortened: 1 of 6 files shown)


jdneo:
This pull request **introduces 3 alerts** when merging 4bb90189cbf02936429b00a2e06c1624d8c1d7cf into bcafa92b04842af12f44e3be37b5b9632c7f0478 - [view on LGTM.com](https://lgtm.com/projects/g/jdneo/vscode-leetcode/rev/pr-db36afa67de63819861b9053eaa81d1da2bae1c1)

**new alerts:**

* 2 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

*Comment posted by [LGTM.com](https://lgtm.com)*

src/webview/leetCodeResultProvider.ts
```diff
@@ -34,19 +40,98 @@ class LeetCodeResultProvider implements Disposable {
...
+                ${body}
             </body>
-            </html>`;
+            </html>
+        `;
     }
 }
 
+// tslint:disable-next-line:max-classes-per-file
+abstract class Result {
```
jdneo: Why not just using interface?

src/webview/leetCodeResultProvider.ts
```diff
@@ -2,29 +2,35 @@
...
+import { MarkdownEngine } from "./MarkdownEngine";
 
 class LeetCodeResultProvider implements Disposable {
 
     private context: ExtensionContext;
     private panel: WebviewPanel | undefined;
+    private mdEngine: MarkdownEngine;
 
     public initialize(context: ExtensionContext): void {
+        this.mdEngine = new MarkdownEngine();
```
jdneo: Having a question here. I believe that we have fully decoupled the markdown engine with the webviews. So do we still need to create new markdown instance in each webview?

src/webview/leetCodeResultProvider.ts
```diff
@@ -34,19 +40,98 @@ class LeetCodeResultProvider implements Disposable {
         }
     }
 
-    private async provideHtmlContent(result: string): Promise<string> {
-        return `<!DOCTYPE html>
-            <html lang="en">
+    private parseResult(raw: string): Result {
+        try {
+            switch (raw[2]) {
+                case "√": {
```
jdneo: I'm not quite sure though.

Be careful about the icon we used here to check the result, because it may be different on different OS, see: https://github.com/skygragon/leetcode-cli/tree/master/icons.

On my Mac, it's `✘` and `✔`

Maybe we need to extract a logic to handle the different symbols from: https://github.com/jdneo/vscode-leetcode/blob/master/src/commands/list.ts#L46-L61

BTW, you can see that this is another limitation by using the `cli` instead of our own `lib`. 😢

src/webview/leetCodeResultProvider.ts
```diff
@@ -34,19 +40,98 @@ class LeetCodeResultProvider implements Disposable {
...
+                `* ${accpet.runtime}`,
+                `* ${accpet.memory}`,
+            ].join("\n"));
+        } else {
+            const failed: FailedResult = result as FailedResult;
+            body = this.mdEngine.render([
+                `## ${result.status}`,
+                `* ${result.passed}`,
+                ``,
+                `### Testcase`, // TODO: add command to copy raw testcase
```
jdneo: Curious, what will be the todo item here?

src/webview/leetCodeResultProvider.ts
```diff
@@ -34,19 +40,98 @@ class LeetCodeResultProvider implements Disposable {
...
+                `\`\`\`\n${failed.testcase}\n\`\`\``,
+                `### Answer`,
+                `\`\`\`\n${failed.answer}\n\`\`\``,
+                `### Expected`,
+                `\`\`\`\n${failed.expected}\n\`\`\``,
```
jdneo: Can we use <br> here for a new line?

Vigilans: It seems that the raw result string always uses "\n" for a new line...e.g. the error messages.

src/webview/leetCodeResultProvider.ts
Vigilans: Copy the raw testcase string for the convenience of submitting with custom testcase.

src/webview/leetCodeResultProvider.ts
Vigilans: I checked the icon in the site, and covered all the relative chars in the regexp in the new commit.

Vigilans:
Currently, the parsing does not cover this case:
image

However, the `[Warn]` line will be parsed as one part of `  √ Your runtime...`, so maybe its visual effect will not be so ugly.

If there is issue about it in the future, further logic can be added at that time.

src/webview/leetCodeResultProvider.ts
jdneo: Got it.

src/webview/leetCodeResultProvider.ts
```diff
@@ -34,19 +37,68 @@ class LeetCodeResultProvider implements Disposable {
         }
     }
 
-    private async provideHtmlContent(result: string): Promise<string> {
-        return `<!DOCTYPE html>
-            <html lang="en">
+    private parseResult(raw: string): IResult {
+        raw = raw.concat("  √ "); // Append a dummy sentinel to the end of raw string
```
jdneo: What will happen if we don't append `√`?

src/webview/leetCodeResultProvider.ts
```diff
@@ -34,19 +37,68 @@ class LeetCodeResultProvider implements Disposable {
...
+        const result: IResult = { messages: [] };
+        let entry: RegExpExecArray | null;
+        do {
+            entry = regSplit.exec(raw);
+            if (!entry) {
+                continue;
+            }
+            const kvMatch: RegExpExecArray | null = regKeyVal.exec(entry[1]);
+            if (kvMatch) {
+                const key: string = kvMatch[1].split("_").map((k: string) => `${k[0].toUpperCase()}${k.slice(1)}`).join(" ");
```
jdneo: Use lodash.startCase to simplify the code. https://lodash.com/docs/4.17.11#startCase 

Same code which can also be replaced: https://github.com/jdneo/vscode-leetcode/blob/f77737d5549761e394f608628c53e317693dd3ab/src/explorer/LeetCodeTreeDataProvider.ts#L220-L222

src/webview/leetCodeResultProvider.ts
Vigilans: The RegEx captures anything in between two `  √ `. If we don't append it, the last section (normally `stdout`) won't be captured.

Worth to note that, with this trick, messages like `[WARN] Failed to get memory percentile` will be captured as one part of `Your runtime beats 81.3 % of python submissions`, and will also be printed out even it does not starts with `  √ `.

jdneo: Hmm... still feeling the logic here is too tricky for maintaince. 

It's ok to merge for now.

Issue #227, opened 2019-03-21
Provide a setting to let user set the Node executable


jdneo:
## 🚀 Feature Proposal

A clear and concise description of what the feature is.

## Motivation

Please outline the motivation for the proposal.

## Contribution

Please describe the solution if you'd like to submit a PR.

PR #233, opened 2019-03-25
Can specify the node path

jdneo:
Resolve: #227 

This can be a solution for those users who use the 3-party shell, like `zshell` on Mac, causing `Node.js` not found.

### DIFF
--- a/src/leetCodeExecutor.ts
+++ b/src/leetCodeExecutor.ts
@@ -11,14 +11,17 @@ import { executeCommand, executeCommandWithProgress } from "./utils/cpUtils";
 import { genFileName } from "./utils/problemUtils";
 import { DialogOptions, openUrl } from "./utils/uiUtils";
 import * as wsl from "./utils/wslUtils";
+import { toWslPath, useWsl } from "./utils/wslUtils";
 
 class LeetCodeExecutor {
     private leetCodeRootPath: string;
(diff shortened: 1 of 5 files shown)


src/leetCodeExecutor.ts
```diff
@@ -11,14 +11,17 @@ import { executeCommand, executeCommandWithProgress } from "./utils/cpUtils";
...
 
 class LeetCodeExecutor {
     private leetCodeRootPath: string;
     private leetCodeRootPathInWsl: string;
+    private nodeExecutable: string;
 
     constructor() {
         this.leetCodeRootPath = path.join(__dirname, "..", "..", "node_modules", "vsc-leetcode-cli");
         this.leetCodeRootPathInWsl = "";
+        this.nodeExecutable = this.getNodePath();
```
Vigilans: Currently, `this.nodeExecutable` is set only once in intialization. If the user changes the `nodePath` config, it won't take effect until reloading. 

Maybe `this.nodeExecutable` could be set as a getter, or set a listener on configuration changed.

jdneo: Add a listener to listen to the configuration change is a good idea. Will update the code.

Vigilans:
Tested without problem on my side (Behavior: when changed the path config and focus out, the settings page will refresh).

PR #235, opened 2019-03-25
Several behavior fix

Vigilans:
## Introduction

Old behavior:
* When quickpicking language, if click other areas to cancel the selection, an `undefined` default language query will pop out:
image
* Sometimes, the leetcode-cli may print useful data, but exit with error code. In this case, an exception will be thrown, and useful data will not be passed.
* When submitting a solution, leetcode-cli may print `Runtime error` and exit with error code. With aforementioned behavior, an exception will be thrown, and result WebView cannot be shown.
image

New behavior:
* Perform extra check before default language query.
* When exiting with error code, collect the result data into the error object.
* When showing result provider, if there is any result data in the caught error object, extract it out to build the WebView. Otherwise, re-throw the error object.

### DIFF
--- a/src/commands/show.ts
+++ b/src/commands/show.ts
@@ -65,7 +65,7 @@ async function fetchProblemLanguage(): Promise<string | undefined> {
     if (defaultLanguage && languages.indexOf(defaultLanguage) < 0) {
         defaultLanguage = undefined;
     }
-    const language: string | undefined = defaultLanguage || await vscode.window.showQuickPick(languages, { placeHolder: "Select the language you want to use" });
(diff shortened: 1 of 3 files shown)


src/commands/show.ts
```diff
@@ -68,7 +68,7 @@ async function fetchProblemLanguage(): Promise<string | undefined> {
     const language: string | undefined = defaultLanguage || await vscode.window.showQuickPick(languages, { placeHolder: "Select the language you want to use" });
```
jdneo: We can also add `ignoreFocusOut: true` into the `showQuickPick()`'s option to avoid user close the drop list by mistake

Vigilans:
Tracked in the issues #236 and #237 .

Issue #203, opened 2019-03-13
command 'leetcode.toggleLeetCodeCn' not found


yanite:
Environment

- *os*: windows 10
- *extension settings*:
- *nodejs version*: v11.11.0
- *vscode version*:  1.32.1
- *extension version*:  0.12.0

安装没问题点击面板上的一切都报这个错误

jdneo:
Please check the document here to see if it could help: https://github.com/jdneo/vscode-leetcode/wiki/%E7%96%91%E9%9A%BE%E8%A7%A3%E7%AD%94#thinking-%E6%8F%90%E7%A4%BA%E9%94%99%E8%AF%AF-command-leetcodexxx-not-found

qianchanger:
我也遇到了.

qianchanger:
I also have this problem......
- mac 10.13.4
- vscode 1.32.3
- nodejs v11.11.0
![image](https://user-images.githubusercontent.com/19400870/54438153-11fa3e80-4771-11e9-968d-a86dd030a9c2.png)

![image](https://user-images.githubusercontent.com/19400870/54438130-0575e600-4771-11e9-8119-6188474361f8.png)

qianchanger:
The above documentation does not seem to contain this problem

qianchanger:
switch this leetcode-plugin to 0.12 version , it works.   :)

jdneo:
Not sure if this is related to this issue: https://github.com/jdneo/vscode-leetcode/issues/201

For example, you are using a third-party bash, like `zshell`, and your node.js is only available in the `zshell`, not the default bash.

Then the extension will not run the leetcode command successfully since it will spawn a child process from the default bash.

Yihemu:
![image](https://user-images.githubusercontent.com/43208747/54474811-ae314d80-4824-11e9-85ab-5283e48e954a.png)
![image](https://user-images.githubusercontent.com/43208747/54474830-d7ea7480-4824-11e9-9a36-bd8d9c125f38.png) 
But i haven't used a third-party bash.

jdneo:
Have you tried this: https://github.com/jdneo/vscode-leetcode/wiki/%E7%96%91%E9%9A%BE%E8%A7%A3%E7%AD%94#thinking-%E6%8F%90%E7%A4%BA%E9%94%99%E8%AF%AF-command-leetcodexxx-not-found

Yihemu:
[...]

PR #249, opened 2019-04-01
Add try-catch block during extesion's activation

Vigilans:
Address https://github.com/jdneo/vscode-leetcode/issues/203#issuecomment-478414326.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -24,7 +24,7 @@ import { leetCodeSolutionProvider } from "./webview/leetCodeSolutionProvider";
 export async function activate(context: vscode.ExtensionContext): Promise<void> {
     try {
         if (!await leetCodeExecutor.meetRequirements()) {
-            throw new Error("The environment doesn't meet requirements");
+            throw new Error("The environment doesn't meet requirements.");
         }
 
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -16,51 +16,57 @@ import { leetCodeChannel } from "./leetCodeChannel";
...
+            vscode.commands.registerCommand("leetcode.submitSolution", (uri?: vscode.Uri) => submit.submitSolution(uri)),
+            vscode.commands.registerCommand("leetcode.switchDefaultLanguage", () => switchDefaultLanguage()),
+        );
 
-    await leetCodeExecutor.switchEndpoint(plugin.getLeetCodeEndpoint());
-    leetCodeManager.getLoginStatus();
+        await leetCodeExecutor.switchEndpoint(plugin.getLeetCodeEndpoint());
+        leetCodeManager.getLoginStatus();
+    } catch (error) {
+        leetCodeChannel.appendLine(error.message);
```
jdneo: `leetCodeChannel.appendLine(error.toString())`

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/toString#Description

PR #260, opened 2019-04-06
Apply vscode's markdown settings to MarkdownEngine

### LINKED ISSUE #246: Let font size of problem describe file equal to editor.fontSize not workbench font size.
## 🚀 Feature Proposal

Let font size of problem describe file equal to editor.fontSize not workbench font size.

## Motivation

The font size of problem describe file is to small.

image

As it opening in editor section why the font size not equal editor font size?

image

Vigilans:
## Introduction

Resolved #246

#### Before
image

#### font size: 14px -> 16px
image

#### font family: prepend 'Consolas'
image

### DIFF
--- a/src/webview/markdownEngine.ts
+++ b/src/webview/markdownEngine.ts
@@ -7,6 +7,7 @@ import * as os from "os";
 import * as path from "path";
 import * as vscode from "vscode";
 import { leetCodeChannel } from "../leetCodeChannel";
+import { isWindows } from "../utils/osUtils";
 
 class MarkdownEngine implements vscode.Disposable {
 
@@ -147,10 +148,18 @@ class MarkdownConfiguration {
 
     public constructor() {
         const markdownConfig: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("markdown");
-        this.fontFamily = markdownConfig.get<string | undefined>("preview.fontFamily", undefined);
-        this.fontSize = Math.max(8, +markdownConfig.get<number>("preview.fontSize", NaN));
-        this.lineHeight = Math.max(0.6, +markdownConfig.get<number>("preview.lineHeight", NaN));
         this.extRoot = path.join(vscode.env.appRoot, "extensions", "markdown-language-features");
+        this.lineHeight = Math.max(0.6, +markdownConfig.get<number>("preview.lineHeight", NaN));
+        this.fontSize = Math.max(8, +markdownConfig.get<number>("preview.fontSize", NaN));
+        this.fontFamily = this.resolveFontFamily(markdownConfig);
+    }
+
+    private resolveFontFamily(config: vscode.WorkspaceConfiguration): string | undefined {
+        let fontFamily: string | undefined = config.get<string | undefined>("preview.fontFamily", undefined);
+        if (isWindows() && fontFamily && fontFamily === config.inspect<string>("preview.fontFamily")!.defaultValue) {
+            fontFamily = `${fontFamily}, 'Microsoft Yahei UI'`;
+        }
+        return fontFamily;
     }
 }
 
--- a/src/webview/markdownEngine.ts
+++ b/src/webview/markdownEngine.ts
@@ -144,7 +144,7 @@ class MarkdownConfiguration {
     public readonly extRoot: string; // root path of vscode built-in markdown extension
     public readonly lineHeight: number;
     public readonly fontSize: number;
-    public readonly fontFamily: string | undefined;
+    public readonly fontFamily: string;
 
     public constructor() {
         const markdownConfig: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("markdown");
@@ -154,7 +154,7 @@ class MarkdownConfiguration {
         this.fontFamily = this.resolveFontFamily(markdownConfig);
     }
 
-    private resolveFontFamily(config: vscode.WorkspaceConfiguration): string | undefined {
+    private resolveFontFamily(config: vscode.WorkspaceConfiguration): string {
         let fontFamily: string = config.get<string>("preview.fontFamily", "");
         if (isWindows() && fontFamily === config.inspect<string>("preview.fontFamily")!.defaultValue) {
             fontFamily = `${fontFamily}, 'Microsoft Yahei UI'`;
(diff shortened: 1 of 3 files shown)


src/webview/markdownEngine.ts
```diff
@@ -107,4 +138,29 @@ class MarkdownEngine {
...
+        this.extRoot = path.join(vscode.env.appRoot, "extensions", "markdown-language-features");
+        this.lineHeight = Math.max(0.6, +markdownConfig.get<number>("preview.lineHeight", NaN));
+        this.fontSize = Math.max(8, +markdownConfig.get<number>("preview.fontSize", NaN));
+        this.fontFamily = this.resolveFontFamily(markdownConfig);
+    }
+
+    private resolveFontFamily(config: vscode.WorkspaceConfiguration): string | undefined {
+        let fontFamily: string | undefined = config.get<string | undefined>("preview.fontFamily", undefined);
+        if (isWindows() && fontFamily && fontFamily === config.inspect<string>("preview.fontFamily")!.defaultValue) {
+            fontFamily = `${fontFamily}, 'Microsoft Yahei UI'`;
```
Vigilans: VS Code in Windows uses SongTi to render Chinese characters. So, if a user didn't change the default `preview.fontFamily` settings, we prepend a YaHei to it to represent Chinese texts.

As for the licensing of the font, please refer to following links:
* https://stackoverflow.com/questions/1805324/how-to-use-web-fonts-legally
* https://www.zhihu.com/question/314779554

Generally speaking, it won't be illegal if we only _reference_ the font, in comparation with _including_ the font by things like `font-face`.

src/webview/markdownEngine.ts
```diff
@@ -7,39 +7,70 @@ import * as os from "os";
...
-    private readonly extRoot: string; // root path of vscode built-in markdown extension
+    private engine: MarkdownIt;
+    private config: MarkdownConfiguration;
+    private listener: vscode.Disposable;
 
     public constructor() {
-        this.engine = this.initEngine();
-        this.extRoot = path.join(vscode.env.appRoot, "extensions", "markdown-language-features");
+        this.reload();
+        this.listener = vscode.workspace.onDidChangeConfiguration((event: vscode.ConfigurationChangeEvent) => {
```
Vigilans: Besides, all the WebViews should register a listener, updating their html when configuration has changed.

To do it, I suggest we abstract a base WebView out of three webviews, and register the listener there.

src/webview/markdownEngine.ts
```diff
@@ -107,4 +138,29 @@ class MarkdownEngine {
...
+    public constructor() {
+        const markdownConfig: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("markdown");
+        this.extRoot = path.join(vscode.env.appRoot, "extensions", "markdown-language-features");
+        this.lineHeight = Math.max(0.6, +markdownConfig.get<number>("preview.lineHeight", NaN));
+        this.fontSize = Math.max(8, +markdownConfig.get<number>("preview.fontSize", NaN));
+        this.fontFamily = this.resolveFontFamily(markdownConfig);
+    }
+
+    private resolveFontFamily(config: vscode.WorkspaceConfiguration): string | undefined {
+        let fontFamily: string | undefined = config.get<string | undefined>("preview.fontFamily", undefined);
```
jdneo: I think just `config.get<string>("preview.fontFamily")` would be fine

src/webview/markdownEngine.ts
```diff
@@ -107,4 +138,29 @@ class MarkdownEngine {
...
+
+    public readonly extRoot: string; // root path of vscode built-in markdown extension
+    public readonly lineHeight: number;
+    public readonly fontSize: number;
+    public readonly fontFamily: string | undefined;
+
+    public constructor() {
+        const markdownConfig: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("markdown");
+        this.extRoot = path.join(vscode.env.appRoot, "extensions", "markdown-language-features");
+        this.lineHeight = Math.max(0.6, +markdownConfig.get<number>("preview.lineHeight", NaN));
```
jdneo: Why we set the default value to `NaN` here?

src/webview/markdownEngine.ts
```diff
@@ -7,39 +7,70 @@ import * as os from "os";
...
+    private config: MarkdownConfiguration;
+    private listener: vscode.Disposable;
 
     public constructor() {
-        this.engine = this.initEngine();
-        this.extRoot = path.join(vscode.env.appRoot, "extensions", "markdown-language-features");
+        this.reload();
+        this.listener = vscode.workspace.onDidChangeConfiguration((event: vscode.ConfigurationChangeEvent) => {
+            if (event.affectsConfiguration("markdown")) {
+                this.reload();
```
jdneo: `reload()` will refresh the configuration the markdown engine. I think we can refresh the webview as well.

src/webview/markdownEngine.ts
Vigilans: This is the strategy adopted by [built-in markdown extension](https://github.com/Microsoft/vscode/blob/master/extensions/markdown-language-features/src/features/previewConfig.ts#L45)

I think the logic is as following:
1. if no config for lineHeight is specified, then the result is `Math.max(0.6, NaN) === NaN`, so [line-height](https://github.com/jdneo/vscode-leetcode/pull/260/files#diff-f7ac20b7043db95594bdb6d86f3ef02bR68) property in css won't be set.
2. if the config is set, we should preserve a minimum line height to 0.6.

src/webview/markdownEngine.ts
jdneo: Yes, feel free to raise another PR after this one.

src/webview/markdownEngine.ts
jdneo: I see...

src/webview/markdownEngine.ts
jdneo: Any thoughts about this one?

Vigilans: Yes, I analysed the code and thought that `undefined` is not needed.

We can just switch to `config.get<string>("preview.fontFamily", "")`.

And when generating css style by config, since `this.config.fontFamily === ""` is falsy, the behavior will still be the same:
```
`    ${this.config.fontFamily ? `font-family: ${this.config.fontFamily};` : ``}`,
```

PR #270, opened 2019-04-14
Extract leetcode webview base class & add md settings listener

Vigilans:
* A webview base class
* webview listener for #260


src/webview/LeetCodeWebview.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public dispose(): void {
+        this.listener.dispose();
+        if (this.panel) {
+            this.panel.dispose();
+        }
+    }
+
+    protected showWebviewInternal(): this is { panel: WebviewPanel } {
```
jdneo: Could you explain more about this method? The return type is `this`, but the actual returned value is `true`

src/webview/leetCodePreviewProvider.ts
```diff
@@ -1,90 +1,44 @@
...
         this.node = node;
-        if (!this.panel) {
-            this.panel = window.createWebviewPanel("leetcode.preview", "Preview Problem", ViewColumn.One, {
-                enableScripts: true,
-                enableCommandUris: true,
-                enableFindWidget: true,
-                retainContextWhenHidden: true,
-                localResourceRoots: markdownEngine.localResourceRoots,
-            });
+        this.description = this.parseDescription(descString, node);
```
jdneo: We can put this line into `getWebviewContent()` to avoid undefined error in the future.

src/webview/leetCodeSolutionProvider.ts
```diff
@@ -1,69 +1,42 @@
...
-            this.panel = window.createWebviewPanel("leetCode.solution", "Top Voted Solution", ViewColumn.Active, {
-                retainContextWhenHidden: true,
-                enableFindWidget: true,
-                localResourceRoots: markdownEngine.localResourceRoots,
-            });
-
-            this.panel.onDidDispose(() => {
-                this.panel = undefined;
-            }, null, this.context.subscriptions);
+        this.solution = this.parseSolution(solutionString);
```
jdneo: Same problem. If in the future, `getWebviewOption()` is called before `this.solution`, then we could get undefined error

src/webview/LeetCodeWebview.ts
```diff
@@ -0,0 +1,70 @@
...
+import { markdownEngine } from "./markdownEngine";
+
+export abstract class LeetCodeWebview implements Disposable {
+
+    protected panel: WebviewPanel | undefined;
+    private context: ExtensionContext;
+    private listener: Disposable;
+
+    public initialize(context: ExtensionContext): void {
+        const { onDidChangeConfiguration } = this.getWebviewOption();
```
jdneo: Why not make `onDidChangeConfiguration` as a class member?

src/webview/LeetCodeWebview.ts
Vigilans: `param is type` is a user-defined [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html) in typescript's grammar. 

It reduced the scope of union type:
```ts
function isTypeDefined(obj: Type | undefined): obj is Type {
    return obj !== undefined;
}

let a: Type | undefined;
a.method() // ts error, a may be undefined

if (isTypeDefined(a)) {
    a.method(); // now a's type is only `Type`, `undefined` is reduced
}
```

`this is type` is a way to guard a class member's type:
https://github.com/Microsoft/TypeScript/issues/26212

`this is { panel: WebviewPanel }` makes sure `this.panel` is not undefined.

So the code here do not need to use `this.panel!.method()` to assert defined:
```ts
if (this.showWebviewInternal()) {
    this.panel.title = `${node.name}: Preview`; // now this.panel's undefined type is reduced
    this.panel.reveal(ViewColumn.One);
}
```

src/webview/leetCodePreviewProvider.ts
Vigilans: The question is, `getWebviewContent()` accepts no arguments (because it will be called in the base class, and there's no way we can know the exact arguments there), so how can we get `descString` and `solutionString`?

I think the proper logic chain is:

---
`getWebviewContent()` binds a html string to `this.panel`
↓
so `getWebviewContent()` will must be called when `this.panel` exists
↓
so `this.panel` should be initialized after `this.description`'s existence is made sure
↓
`this.panel`'s only way to be initialized is by calling `showWebviewInternal()`
↓
so when `this.description` is set before `showWebviewInternal()`, `getWebviewContent()` will never results in undefined error.

---

This is exactly what following code does:
```ts
this.description = this.parseDescription(await leetCodeExecutor.getDescription(node), node);
this.node = node;
if (this.showWebviewInternal()) {
    this.panel.title = `${node.name}: Preview`;
    this.panel.reveal(ViewColumn.One);
}
```

So, the only way to cause such problem is when a new way to initialize `this.panel` is added. Then, it is the adder to ensure `this.panel` is initialized when all it needs is set.

src/webview/LeetCodeWebview.ts
Vigilans: I wanted to make the way to implement two event callback consistent. After some investigation on js overriden mechanism, now I made sure all two methods could be set as class member.

src/webview/LeetCodeWebview.ts
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    protected async onDidChangeConfiguration(event: ConfigurationChangeEvent): Promise<void> {
+        if (this.panel && event.affectsConfiguration("markdown")) {
+            this.panel.webview.html = this.getWebviewContent();
+        }
+    }
+
+    protected async onDidReceiveMessage(/* message */_: any): Promise<void> { /* no special rule */ }
```
jdneo: `/* message */_` -> `_message`

src/webview/LeetCodeWebview.ts
```diff
@@ -0,0 +1,66 @@
...
+        if (!this.panel) {
+            this.panel = window.createWebviewPanel(viewType, title, viewColumn || ViewColumn.Active, {
+                enableScripts: true,
+                enableCommandUris: true,
+                enableFindWidget: true,
+                retainContextWhenHidden: true,
+                localResourceRoots: markdownEngine.localResourceRoots,
+            });
+            this.panel.onDidDispose(this.onDidDisposeWebview, this, this.context.subscriptions);
+            this.panel.webview.onDidReceiveMessage(this.onDidReceiveMessage, this, this.context.subscriptions);
```
jdneo: nit: The `onDidReceiveMessage` can also be disposed in `onDidDisposeWebview` 

> Put it int `subscriptions` is fine though.

src/webview/LeetCodeWebview.ts
jdneo: Then why we are not just calling `showWebviewInternal()` but put it in the if block.

I do not see any benefits to do this.

src/webview/leetCodePreviewProvider.ts
jdneo: I see. Let me think more about this. Just leave the current implementation without change.

src/webview/LeetCodeWebview.ts
Vigilans: tslint reports 
`variable name must be in lowerCamelCase, PascalCase or UPPER_CASE (variable-name)tslint(1) `

src/webview/LeetCodeWebview.ts
Vigilans: another alternative is to pass `this.context.subscriptions` to `workspace.onDidChangeConfiguration`, and do not maintain `configListener` in `LeetCodeWebview`?

src/webview/LeetCodeWebview.ts
Vigilans: You mean
```ts
this.showWebviewInternal();
this.panel!.title = `${node.name}: Preview`; // this.panel's undefined type is not reduced
this.panel!.reveal(ViewColumn.One);
```
?

src/webview/LeetCodeWebview.ts
jdneo: Emm, we can update the tslint.json.

Add `"allow-leading-underscore"` in `variable-name`

src/webview/LeetCodeWebview.ts
jdneo: Just my personal opinion: putting into `subscriptions` is a passive way to clean up the memory:

> An array to which disposables can be added. When this extension is deactivated the disposables will be disposed.

It's more suitable for the command, global objects, etc. Since once the command is registered, in most cases you will not remove it during the extension is running.

While here, once the panel is disposed. We need to dispose the listener as well to avoid that multiple listeners are working in the meantime. So disposing them in the `onDidDisposeWebview()` could be better.

src/webview/LeetCodeWebview.ts
jdneo: The main problem here is, we should extract the fields in `ILeetCodeWebviewOption` as the member of the base class. The reason for doing that:

1. Each web view has its own unique `viewType`
2. Each web view has its own location to display - the `viewColumn `
3. Each web view has the member title to determine the title of the editor.

So after the extraction, what you need to do is:
1. Move all the logic about the panel update into `showWebviewInternal()`
2. Update the title field when needed.

```typescript
this.title = xxxx
this.showWebviewInternal();
```

Vigilans: Ah...the next PR I prepared yesterday has done the work (panel update in `showWebviewInternal`) in another way:

In `showWebviewInternal()`:
```ts
if (!this.panel) {
    // ...old initialization code
} else {
    this.panel.title = title;
    if (viewColumn === ViewColumn.Two) {
        // Make sure second group exists. See vscode#71608 issue
        commands.executeCommand("workbench.action.focusSecondEditorGroup").then(() => {
            this.panel!.reveal(viewColumn, preserveFocus);
        });
    } else {
        this.panel.reveal(viewColumn, preserveFocus);
    }
}
```

In `show(...)`:
```ts
public async show(node: IProblem, isSideMode: boolean = false): Promise<void> {
    this.description = this.parseDescription(await leetCodeExecutor.getDescription(node), node);
    this.node = node;
    this.sideMode = isSideMode;
    this.showWebviewInternal();
    if (this.sideMode) {
        // Hide side bar for better view area
        await commands.executeCommand("workbench.action.focusSideBar").then(() => {
            commands.executeCommand("workbench.action.toggleSidebarVisibility");
        });
    }
}
```

In `getWebviewOption()`:
```ts
protected getWebviewOption(): ILeetCodeWebviewOption {
    if (!this.sideMode) { // a state when all the leetcode panels are grouped in Column 2
        return {
            viewType: "leetcode.preview",
            title: `${this.node.name}: Preview`,
            viewColumn: ViewColumn.One,
            preserveFocus: true,
        };
    } else {
        return {
            viewType: "leetcode.preview",
            title: "Description",
            viewColumn: ViewColumn.Two,
        };
    }
}
```
In this way, the data needed to update the panel is dynamically organized in `getWebviewOption()`.

src/webview/LeetCodeWebview.ts
Vigilans: This way?
```ts
private listeners: Disposable[] = [];

//...

this.panel.onDidDispose(this.onDidDisposeWebview, this, this.listeners);
this.panel.webview.onDidReceiveMessage(this.onDidReceiveMessage, this, this.listeners);
workspace.onDidChangeConfiguration(this.onDidChangeConfiguration, this, this.listeners);

//...

protected onDidDisposeWebview(): void {
    this.panel = undefined;
    for (const listener of this.listeners) {
        listener.dispose();
    }
    this.listeners = [];
}

```

In this way, `this.context` will report `never used` by tslint. We could make such modification:
```ts
public initialize(context: ExtensionContext): void {
    this.context = context;
    this.context.subscriptions.push(this);
}
```
It separates the intialization code in `activate` function.

jdneo: just move `this.context` out of webview looks good to me.

src/webview/LeetCodeWebview.ts
jdneo: Do you mean you want to add a new setting called `isSideMode` to determine the location of the webviews?

Vigilans: This is the focus of next PR, which uses column 2 to provide information when writing solution.

jdneo: So far, I cannot get the point of the method `getWebviewOption()` who else will call it?

Vigilans: `getWebviewOption()` is only used in `showWebviewInternal()`, to initialized or update the panel.

jdneo: Then why just make them as the members of the class? From the code flow, the `title` is different according to the different value of `sideMode`. Actually, there is no relationship between them. The title should always update to a more meaningful value no matter where will the webview be located. That's the reason I prefer to use class member instead of `getWebviewOption()`.

Vigilans: I've brought everything except `sideMode` in next PR to this PR. It includes the improvement on the webview base class, like mentioned above.

Vigilans: Do you really think there is no relationship? Did you ever test yourself what is the visual effect when the title keeps `A so long long long long long name: preview` and `A so long long long long long name: solution` in Column 2? Will the user be able to easily change between the tabs?

Why do you pay more attention to something in the next PR which I haven't given illustration yet?

Vigilans: I just want leetcode tabs in column 2 look like this way:
image

Otherwise, in my small-screen PC, a `So long long long a name: Preview` just occupies all the room in the Column 2.

jdneo: Of cause I need to understand what you will do in the next PR since they have some relationship. This will determine whether we need to keep `getWebviewOption()` or make them as the class member.

The codes written in a good way should not only have good experience but also good at self-explain. This is very important to attract more community contribution. That's why I'm keeping asking you those questions. From the code itself, I cannot get your point. There's no GIF/picture to illustrate that. Then how could I know your point is that the tab is so long that make the experience bad?

Besides, From my understanding, the reason to rename to the title to `Description` is that you hope the size of the tab is not too big when VS Code has two columns. The real reason actually is having limited resolution, and multiple columns is one of the possible reason to cause the limited resolution.

Take the VS Code embedded Markdown extension as an example, You can see there is no optimization when the tab is too long.
image

jdneo: I'm not saying that we should not do the optimation. It's acceptable to do it as long as the code logic is easy to understand.

Vigilans:
illustration about side mode:

Walkthrough:
WalkThrough

ShowProblem command also benefits from it:
ShowProblem

Solution weview always follows the state of preview:
Preview   Solution

What if the title doesn't change (this is on a wide-screen computer):
WhatIf

Vigilans:
My argument for using `getWebviewOption()` is that it is in parallel with `getWebviewContent()`:

1. They all provide the function of dynamic updating. 
Now that `viewColumn` and `title`'s need to be updated dynamiclly is recognized, we can view `getWebviewOption()` and `getWebviewContent()` as two decoupled procedures of updating a webview. 
`show()` is only a stub receiving input parameters, followed with `showWebviewInternal()`.

2.  `.title`, `.viewColumn` & `.webivew.html` should be three equal-level buiding-blocks of a webview. 
If we keep `title` as a member of base class, then the same applies for `.webview.html`.

3. As a class member, `.title`, `.viewColumn` & `.webivew.html` has already been saved in `this.panel`, we don't need to keep another copy of them.

PR #271, opened 2019-04-16
Add side mode to webview

Vigilans:
Walkthrough:
WalkThrough

ShowProblem command also benefits from it:
ShowProblem

Solution weview always follows the state of preview:
Preview   Solution

What if the title doesn't change (this is on a wide-screen computer):
WhatIf


src/webview/leetCodePreviewProvider.ts
```diff
@@ -11,19 +11,40 @@ class LeetCodePreviewProvider extends LeetCodeWebview {
...
+    }
+
+    public async show(node: IProblem, isSideMode: boolean = false): Promise<void> {
         this.description = this.parseDescription(await leetCodeExecutor.getDescription(node), node);
         this.node = node;
+        this.sideMode = isSideMode;
         this.showWebviewInternal();
+        if (this.sideMode) {
+            // Hide the side bar for better view area
+            commands.executeCommand("workbench.action.focusSideBar").then(() => {
```
Vigilans: This is not an elegant workaround...
But I searched for hours the api to check the side bar status but in vain...

Workarounds which would have been possible:
1. api in `vscode.window` namespace. We could control the visible state of things like terminal/text editor, but side bar isn't in the list.
2. context value. One [issue](https://github.com/Microsoft/vscode/issues/66956) in vscode suggests that `sidebarVisible` can be used for `toggleSideBar` to work like `closeSidebar`. Yet neither can I find out how to invoke `executeCommand()` with `when` context, nor can I retrieve the context value (there is a command named `setContext`, yet `getContext` does not exist)

jdneo:
Thanks, I'll take a look tomorrow.

src/webview/LeetCodeWebview.ts
```diff
@@ -30,7 +30,14 @@ export abstract class LeetCodeWebview implements Disposable {
             workspace.onDidChangeConfiguration(this.onDidChangeConfiguration, this, this.listeners);
         } else {
             this.panel.title = title;
-            this.panel.reveal(viewColumn, preserveFocus);
+            if (viewColumn === ViewColumn.Two) {
+                // Make sure second group exists. See vscode#71608 issue
+                commands.executeCommand("workbench.action.focusSecondEditorGroup").then(() => {
```
jdneo: Do we need `await` here?

src/webview/leetCodePreviewProvider.ts
jdneo: `await` here?

src/webview/leetCodePreviewProvider.ts
```diff
@@ -11,19 +11,40 @@ class LeetCodePreviewProvider extends LeetCodeWebview {
...
 
     protected getWebviewOption(): ILeetCodeWebviewOption {
-        return {
-            viewType: "leetcode.preview",
-            title: `${this.node.name}: Preview`,
-            viewColumn: ViewColumn.One,
-        };
+        if (!this.sideMode) {
+            return {
+                viewType: "leetcode.preview",
```
jdneo: Can we make the viewType as a member of the webview?

src/commands/show.ts
```diff
@@ -111,7 +112,10 @@ async function showProblemInternal(node: IProblem): Promise<void> {
 
         const originFilePath: string = await leetCodeExecutor.showProblem(node, language, outDir);
         const filePath: string = wsl.useWsl() ? await wsl.toWinPath(originFilePath) : originFilePath;
-        await vscode.window.showTextDocument(vscode.Uri.file(filePath), { preview: false, viewColumn: vscode.ViewColumn.One });
+        await Promise.all([
+            vscode.window.showTextDocument(vscode.Uri.file(filePath), { preview: false, viewColumn: vscode.ViewColumn.One }),
+            leetCodePreviewProvider.show(node, true),
```
jdneo: What if we provide an extension setting about the `sideMode` and default is `true`?

src/webview/leetCodePreviewProvider.ts
Vigilans: Yes, we can make `viewType` and `iconPath` both inheritable members of webview.

jdneo: @Vigilans Will the change happen in 0.14.1?

Vigilans: You mean add `iconPath` in 0.14.1?

src/webview/LeetCodeWebview.ts
Vigilans: `showWebviewInternal` is not an aysnc function, and the workaround is not needed anymore when the issue mentioned is solved. So I think it not worthful to make it `await`.

Besides, `this.panel.reveal` is just a RPC call and do not wait for the result (investigated in the issue), so it is also not needed to wait for the command.

src/commands/show.ts
Vigilans: √

src/webview/leetCodePreviewProvider.ts
Vigilans: I think `WebviewProvider.show` should not be an async function in the first place: every operation that needs to await something should be did before calling `show` ---- what is passed to `show` should be the resources prepared. 

The commands are also operations that don't need to be waited, just like `this.panel.reveal` does not wait for the panel to be revealed.

So I pushed a commit to move `getDescription` to `commands/show.ts/previewProblem`, and make `show` not an async function...

Vigilans:
`enableSideMode` walkthrough:
enableSideMode

package.json
```diff
@@ -301,6 +301,12 @@
                         "scope": "application",
                         "description": "Show the submit and test shortcuts in editor or not."
                     },
+                    "leetcode.enableSideMode": {
+                        "type": "boolean",
+                        "default": true,
+                        "scope": "application",
+                        "description": "Determine whether to group all webview pages into Column 2 when solving problems."
```
jdneo: `Column 2` -> `The second editor column`

src/commands/show.ts
```diff
@@ -111,7 +117,10 @@ async function showProblemInternal(node: IProblem): Promise<void> {
 
         const originFilePath: string = await leetCodeExecutor.showProblem(node, language, outDir);
         const filePath: string = wsl.useWsl() ? await wsl.toWinPath(originFilePath) : originFilePath;
-        await vscode.window.showTextDocument(vscode.Uri.file(filePath), { preview: false, viewColumn: vscode.ViewColumn.One });
+        await Promise.all([
+            vscode.window.showTextDocument(vscode.Uri.file(filePath), { preview: false, viewColumn: vscode.ViewColumn.One }),
+            leetCodeConfig.get<boolean>("enableSideMode", true) ? previewProblem(node, true) : Promise.resolve(),
```
jdneo: Why we do nothing when `enableSideMode` is false.

src/webview/leetCodeSolutionProvider.ts
```diff
@@ -3,24 +3,35 @@
...
 class LeetCodeSolutionProvider extends LeetCodeWebview {
 
+    protected readonly viewType: string = "leetcode.solution";
     private solution: Solution;
+    private sideMode: boolean = false;
 
-    public async show(solutionString: string, problem: IProblem): Promise<void> {
+    public show(solutionString: string, problem: IProblem): void {
         this.solution = this.parseSolution(solutionString, problem);
+        this.sideMode = leetCodePreviewProvider.isSideMode();
```
jdneo: Please always fetch the setting `enableSideMode` instead of getting it from the preview Provider.

src/webview/leetCodePreviewProvider.ts
jdneo: Wrap the following code into a function like `hideLeftSideBar()`, then no need to add comments
```
commands.executeCommand("workbench.action.focusSideBar").then(() => {commands.executeCommand("workbench.action.toggleSidebarVisibility");
```

src/webview/leetCodeSolutionProvider.ts
Vigilans: This is not the case: solution provider always follows the state of preview provider, rather than whether side mode is enabled.

If `enableSideMode` is true, but preview webview is not in side mode, then solution

src/commands/show.ts
Vigilans: When opening a preview, clicking `code now`, the preview and text document will be both in Column 1 if `enableSideMode` is set to false. 

If we do `previewProblem` here, the result demo:
demo

package.json
Vigilans: Updated.

src/webview/leetCodePreviewProvider.ts
Vigilans: Updated

src/commands/show.ts
jdneo: Oh, I got your point.

What to do when sideMode is enabled is:

1. Move the preview to 2nd column
2. show the source code file

What to do when sideMode is disabled is:
1. show the source code file

The method name `previewProblem()` is confusing here. I think we should wrap it with a method, named like `previewOnTheSideIfNeeded()`

src/webview/leetCodeSolutionProvider.ts
jdneo: Ok, if we want to make sure `solution` and `result` are always show in the same column, then I think we should directly call

``` js
if (leetCodePreviewProvider.isSideMode()) {
    ...
}
```

instead of they owning their own `sideMode` field.

src/commands/show.ts
Vigilans: Updated (`movePreviewAsideIfNeeded`)

src/webview/leetCodeSolutionProvider.ts
Vigilans: updated

Vigilans:
Document added.

Issue #146, opened 2019-02-21
could provide an option to hide the `Submit to LeetCode` button in the editor?


AEPKILL:
I always accidentally click it when editing the file

taoweicn:
+1
And I think it's better to replace `Submit` with `Test`.

poppinlp:
Agree with it. And a switch in settings too. :)

Eskibear:
My suggestion is to 1) simply add a switch for the code lens, 2) keep both "Submit" and "Test" if  codelens is turned on. 

The config is: 
```json
"leetcode.showShortcuts": {
    "type": "boolean",
    "default": true
}
```

It looks like:
```
...
piece of code
...

Submit | Test
```

Issue #269, opened 2019-04-10
Add  a "show top solution" item to the right click menu, pls


YorksonChang:
## 🚀 Feature Proposal

When right click on the editor window，show the  "show the top solution" option in the menu.
Then users needn't look up the problem on the left to get the solution.

## Motivation

直接在代码窗口右键即可查看本问题的最优解法，而不需要在左栏重新找问题然后找答案。

## Contribution

Please describe the solution if you'd like to submit a PR.

jdneo:
Actions to take:

1. Add it into the context menu
2. Add it into the code lens shortcuts

PR #273, opened 2019-04-16
Can config whether to show or hide the codelens

jdneo:
Resolve: #146


Vigilans:
#269

Also, add a `solution` button to codelens?

jdneo:
Good idea. I'll do it in another PR.

PR #278, opened 2019-04-16
Resolve #261

### LINKED ISSUE #261: Solution webview doesn't decode unicode characters
## 🐛 Bug Report

image

\u2264 should have been ≤.

Vigilans:
\u2264 is ≤ now.
image

Fix #261


src/commands/show.ts
```diff
@@ -50,10 +50,11 @@ export async function showSolution(node?: LeetCodeNode): Promise<void> {
     }
     try {
         let solution: string = await leetCodeExecutor.showSolution(node, language);
-        // remove backslash in espaced \'...\'(generated by leetcode's database)
-        solution = solution.replace(/\\'/g, "'");
+        // tslint:disable-next-line: no-eval
+        solution = eval(`"${solution.replace(/\n/g, "\\n")}"`); // decode solution string with escaped characters
```
Vigilans: `eval` is used to decode the escaped characters with javascript engine;

`solution.replace(/\n/g, "\\n")` is because the original plugin author only replaces `\\n` with `\n', but left other characters undecoded.

jdneo: `eval()` is not recommended to use, see:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#Do_not_ever_use_eval!
- https://stackoverflow.com/a/10474424

Do we have other alternatives?

Vigilans: There is, using JSON.parse + decodeURIComponent:
* https://www.quora.com/What-s-quick-way-to-unescape-unicode-escaped-strings-such-as-uXXXX-etc-in-JavaScript-or-other-language
* https://stackoverflow.com/questions/7885096/how-do-i-decode-a-string-with-escaped-unicode

However, after tested on some test case, I think the solution string is not designed for using `JSON.parse`.

I did following work to make https://leetcode.com/problems/regular-expression-matching/discuss/5684 work:
1. `replace(/\n/g, "\\n")`: this is the fault with leetcode-plugin author, also needed for `eval`.
2. `replace(/"/g, '\\"')`: also needed for `eval`.
3. `replace(/\\'/g, "'")`: convert `\'` back to `'` because it doesn't recognize escaped `\'`.
4. `replace(/\t/g, "\\t")`: non-escaped indent is not recognized (the opposite of case 3).

According to case 3, 4, there may be more corner cases with escaping in JSON. So, I think the solution's `content` property in leetcode was originally designed to generate raw javascript code file, and if we want to use it as string, `eval` may be a proper choice.

jdneo:
This pull request **introduces 1 alert** when merging c25bc62cc0bb305c8115d5d3c19892c84c83e519 into b90354e2212de2104fffec637bd24aa2539ad972 - [view on LGTM.com](https://lgtm.com/projects/g/jdneo/vscode-leetcode/rev/pr-7b9db62d5a203e75422a8ade966fe2327eec8368)

**new alerts:**

* 1 for Incomplete string escaping or encoding

---

*Comment posted by [LGTM.com](https://lgtm.com)*

PR #289, opened 2019-04-23
Add `showCommentDescription` config

Vigilans:
Defaults to false.

Relies on https://github.com/leetcode-tools/leetcode-cli/pull/11

Solves #287 

Demo:
demo2

Issue #287, opened 2019-04-22
0.14.0 problem description


donifer:
## 🐛 Bug Report
Before 0.14.0 when I started a new problem the instructions would be in the new problem file commented out. But now I don't get the description in the problem's comments. I just get something like this without the description
```js
/*
 * @lc app=leetcode id=35 lang=javascript
 *
 * [35] Search Insert Position
 */
/**
 * @param {number[]} nums
 * @param {number} target
 * @return {number}
 */
var searchInsert = function(nums, target) {
    
};


```

## To Reproduce
Start any problem from the sidenav.

## Expected behavior
The problems should include the description.

## Extension Output
```
/*
 * @lc app=leetcode id=35 lang=javascript
 *
 * [35] Search Insert Position
 */
/**
 * @param {number[]} nums
 * @param {number} target
 * @return {number}
 */
var searchInsert = function(nums, target) {
    
};

[35] Search Insert Position  

https://leetcode.com/problems/search-insert-position/description/

Tags:   algorithms   array   binary-search 

Langs:  c   cpp   csharp   golang   java   javascript   kotlin   php   python   python3   ruby   rust   scala   swift 

* algorithms
* Easy (40.67%)
* Likes:    1225
* Dislikes: 170
* Total Accepted:    386.1K
* Total Submissions: 948.8K
* Testcase Example:  '[1,3,5,6]\n5'

<p>Given a sorted array and a target value, return the index if the target is found. If not, return the index where it would be if it were inserted in order.</p>

<p>You may assume no duplicates in the array.</p>

<p><strong>Example 1:</strong></p>

<pre>
<strong>Input:</strong> [1,3,5,6], 5
<strong>Output:</strong> 2
</pre>

<p><strong>Example 2:</strong></p>

<pre>
<strong>Input:</strong> [1,3,5,6], 2
<strong>Output:</strong> 1
</pre>

<p><strong>Example 3:</strong></p>

<pre>
<strong>Input:</strong> [1,3,5,6], 7
<strong>Output:</strong> 4
</pre>

<p><strong>Example 4:</strong></p>

<pre>
<strong>Input:</strong> [1,3,5,6], 0
<strong>Output:</strong> 0
</pre>

```

## Your Environment

- *os*: macOS
- *extension settings*: 
```json
{
  "leetcode.defaultLanguage": "javascript",
  "leetcode.showLocked": true,
  "leetcode.enableStatusBar": false,
  "leetcode.enableShortcuts": false
}
[...]

PR #294, opened 2019-04-26
Support showing description from text document

Vigilans:
Solves #286

Relies on https://github.com/leetcode-tools/leetcode-cli/pull/13

Demonstration:
demo2


jdneo:
@Vigilans First of all, thank you for the enthusiasm you give.

One thing I have to mentions is that, please do not try to solve multiple issues in one PR. This very hard for tracking the issues.

Could you please revert this PR to make it only solve #287 first. And please let me know what I can do to help you achieve that.

Thanks.

Vigilans:
#287 has already been solved in #289 4 days ago and wait for the `leetcode-cli` dependency update.

This three issues all point to the same thing(comment in #287 is duplicate of #286), and there is no meaning in splitting them: it will only result in a PR of 90% content and a PR with 10% content.

Vigilans:
> Could you please revert this PR to make it only solve #287 first. And please let me know what I can do to help you achieve that.

This PR doesn't solve #287

jdneo:
I see. I've added comments in #289.

But still, I would like to request you to split the PR even it's 10% + 90%. Thank you for your understanding.

PR #313, opened 2019-05-08
feat: Add 'show solution' in editor context menu & code lens

### LINKED ISSUE #269: Add  a "show top solution" item to the right click menu, pls

jdneo:
Resolve #269 

demo

Need to release cli first.

### DIFF
--- a/src/commands/show.ts
+++ b/src/commands/show.ts
@@ -17,9 +17,9 @@ import { leetCodePreviewProvider } from "../webview/leetCodePreviewProvider";
 import { leetCodeSolutionProvider } from "../webview/leetCodeSolutionProvider";
 import * as list from "./list";
 
-export async function previewProblem(node: IProblem): Promise<void> {
+export async function previewProblem(node: IProblem, isSideMode: boolean = false): Promise<void> {
     const descString: string = await leetCodeExecutor.getDescription(node);
-    leetCodePreviewProvider.show(descString, node);
+    leetCodePreviewProvider.show(descString, node, isSideMode);
 }
 
 export async function showProblem(node?: LeetCodeNode): Promise<void> {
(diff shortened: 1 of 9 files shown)


src/commands/show.ts
```diff
@@ -47,17 +47,24 @@ export async function searchProblem(): Promise<void> {
...
+    if (input instanceof LeetCodeNode) {
+        problemInput = input.id;
+    } else if (input instanceof vscode.Uri) {
+        problemInput = `"${input.fsPath}"`;
+    } else {
+        vscode.window.showErrorMessage("Invalid input to fetch the solution data");
         return;
     }
+
     const language: string | undefined = await fetchProblemLanguage();
```
Vigilans: We could utilize file extension (`path.extName()`) and `shared.ts/langExt` to determine the language, thus no need for re-selecting the language when showing solution from the code:
```ts
export const langExt: Map<string, string> = new Map([
    ["bash", "sh"],
    ["c", "c"],
    ["cpp", "cpp"],
    ["csharp", "cs"],
    ["golang", "go"],
    ["java", "java"],
    ["javascript", "js"],
    ["kotlin", "kt"],
    ["mysql", "sql"],
    ["php", "php"],
    ["python", "py"],
    ["python3", "py"],
    ["ruby", "rb"],
    ["rust", "rs"],
    ["scala", "scala"],
    ["swift", "swift"],
]);
```
(Yet this choice comes with a risk: currently showing solution for languages like scala has a high probability  to fail. This feature could be added in later PR when logic of solution plugin is improved.)

jdneo: another rick is that it's hard to differentiate py3 & py2 from the `extName`

PR #380, opened 2019-08-02
implement leetcode.relativeFilePath

magic-akari:


package.json
```diff
@@ -319,7 +320,248 @@
...
+                                        "type": "string"
+                                    },
+                                    "filename": {
+                                        "type": "string"
+                                    }
+                                },
+                                "minProperties": 1
+                            }
+                        },
+                        "additionalProperties": {
```
jdneo: Hmm, I'm thinking that when will the user to set additional properties here?

package.json
```diff
@@ -319,7 +320,248 @@
                     "leetcode.outputFolder": {
                         "type": "string",
                         "scope": "application",
-                        "description": "The relative path to save the problem files."
+                        "description": "[Deprecated] The output folder to save the problem files."
+                    },
+                    "leetcode.output": {
```
jdneo: So far I like this schema!

jdneo:
Hi @magic-akari,

Apologize for the late reply about this PR. I was busying with other stuff the days before. Today I just got a chance to take a look at it. Very nicely implemented I think. Just some small questions. Would you mind to take a look?

package.json
jdneo: How about:
```
leetcode.filepath: {
    folder: xxx
    fileName: xxx
}
```

package.json
jdneo: How about:
```
leetcode.filepath: {
    folder: xxx
    fileName: xxx
}
```

package.json
magic-akari: When leetcode supports new languages and we haven't updated them yet, `additionalProperties ` can be used.

package.json
magic-akari: done

magic-akari:
There are a lot of duplicate codes in `leetcode.filePath` config.
I have tried to use [json schema's ref to reuse them](https://json-schema.org/understanding-json-schema/structuring.html#reuse), but it doesn't work in vscode.
Do you have any idea?

jdneo:
@magic-akari Yes, from engineering aspect, have the configuration clean and neat is a worth considering topic. While from the feature aspect, the changes here is good to go. Let me take a look at how to simplify the configuration in `package.json` this weekend.

If we cannot address it for now, we can create an issue to track that and ship this feature out to benefit our users first.

Thanks.

jdneo:
Thank you @magic-akari! Nice work!

Issue #401, opened 2019-08-27
Update the extension icon to align with the new VS Code design


jdneo:
The icon we are using now is not align with the VS Code new design.
![icon](https://raw.githubusercontent.com/jdneo/vscode-leetcode/master/resources/LeetCode.png)

poppinlp:
I've asked my friend for the logo.
Actually, he has asked me if there are some UI/UX design which we need, after he gave me that sidebar icon.

jdneo:
That would be great if he is willing to help. So far the extension icon and the sidebar icon is enough I think.

Does he has a GitHub account? We should thank to his contribution in our document.

poppinlp:
![image](https://user-images.githubusercontent.com/1970625/64581808-1f118680-d3be-11e9-8220-26ceb58132eb.png)
@jdneo We get 2 versions of the icon. The first one is closer to the new vscode icon. The second one is closer to the leetcode icon. Which one do you prefer?

jdneo:
Thanks @poppinlp. I prefer the first one, what do you think?

poppinlp:
Yep. Me too.
Okey. I'll send PR then. And adding him in the `Acknowledgement` part for his contribution.

jdneo:
Can't wait for it!! 😄

jdneo:
Hey @poppinlp ,

Any updates for the new icon PR? I'm planning to make a new release recently and wish we can have the new icon in the new release. 😄

poppinlp:
Sorry for the late update. I've send a PR for it.
The original svg image got something wrong, so I asked my friend to give me a new one this morning.

jdneo:
That's fine. Don't worry.

PR #419, opened 2019-09-16
chore: replace the LeetCode.svg and LeetCode.png

poppinlp:
- Replace the `LeetCode.svg` and `LeetCode.png`
- Add JIEJIAN21 in contributors
- Seems my markdown plugin reformated the table in README.md to make the separator in the same column.

Address #401


jdneo:
Thanks @poppinlp, could you fix the CI error? 

`ERROR  Images in README.md must come from an HTTPS source: ./resources/LeetCode.png`

poppinlp:
This error is occured by the url of `LeetCode.png` in `README.md` and `README_zh-CN.md`.
I've modified the url to local relative path instead of the file from master branch of github repository.

Actually, I think it's good to use relative path here since we should use the resources from current branch. But seems vsce dose not support it right now according to [this issue](https://github.com/Microsoft/vscode-vsce/issues/192).

I've changed it back to pass the travis CI.

jdneo:
Hey @poppinlp

Just one more question. Now the extension icon and the activity bar icon are both using the new picture. Are we mean to discard the activity bar icon that your friend designed just before?

poppinlp:
I think I made a mistake. I should only replace the image in README. The previous activity icon is much more fits the sidebar. I'll bring it back now.
image

PR #426, opened 2019-09-18
fix: CodeLens position

magic-akari:


src/codelens/CustomCodeLensProvider.ts
```diff
@@ -22,23 +22,16 @@ export class CustomCodeLensProvider implements vscode.CodeLensProvider {
...
         if (!matchResult) {
             return undefined;
         }
 
-        let codeLensLine: number = content.length - 1;
-
-        for (let i: number = 0; i < content.length; ++i) {
-            if (content[i].indexOf("@lc code=end") >= 0) {
+        let codeLensLine: number = document.lineCount - 1;
+        for (let i: number = document.lineCount - 1; i >= 0; i--) {
```
magic-akari: It seems that the algorithm of the `@lc code=end` line is different from the algorithm in [leetcode-cli](https://github.com/leetcode-tools/leetcode-cli/blob/f5c6f9aa7a79f6849a985b06839c81be1fb968bb/lib/file.js#L120).

jdneo: Any concern for this?

magic-akari: If there are two lines which contain `@lc code=end`, We will get inconsistent behavior.
But it is rare, so we don't have to deal with it.

jdneo: Since we have no context information, so this is the case that we cannot deal with whatever algorithm we take. So let's just keep it as it is.

Thank you for the feedback anyway.

PR #476, opened 2019-11-15
FEATURE: Display 'active session' in the status bar

malfusion:
_**FEATURE: Display 'active session' in the status bar**_

The logged in users's username is currently shown in the status bar. However, I needed the ability to check which session I was currently in before submitting my solutions. It was inconvenient to keep using the "Manage sessions" option to check the currently active session. So I built one for my own use case. 
I would like to open a pull request for this feature. If you find this feature useful, let me know your changes/suggestions, and do consider merging this feature to the extension.

_**What it does:**_
- The proposed implementation builds on top of 'listSessions' functionality, to obtain the currently active session.
- The extension retrieves the current session every 10 seconds and updates the status bar with the session name as well

_**Possible changes:**_
- Polling interval can be made to be user-defined in settings

_**Result - Status bar**_
image

### DIFF
--- a/src/commands/session.ts
+++ b/src/commands/session.ts
@@ -54,6 +54,17 @@ export async function manageSessions(): Promise<void> {
     }
 }
 
+export async function getActiveSession(): Promise<ISession | void> {
+    try {
+        const sessions: ISession[] = await getSessionList();
+        const activeSession: Array<ISession> = sessions.filter((s: ISession) => s.active);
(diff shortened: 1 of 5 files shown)


jdneo:
Hi @malfusion,

Thank you for your contribution. Apologize for my late reply.

One thing I'm not quite understand is that why we need to poll the active session? In case the active session is changed by other tools?

PR #487, opened 2019-12-03
add leetcode.signinByCookie but simple change

yihong0618:
This is very simple change maybe for  short-term  solution for #478 
Sorry for the first time write ts.
Please help me review~
@jdneo 
Thanks a lot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@
 
 - You can also use the following command to sign in/cookie in/out:
   - **LeetCode: Sign in**
-  - **LeetCode: Cookie in**
+  - **LeetCode: Sign in (by cookie)**
   - **LeetCode: Sign out**
 
 ---
--- a/README.md
+++ b/README.md
@@ -40,8 +40,9 @@
 
 - Simply click `Sign in to LeetCode` in the `LeetCode Explorer` will let you **sign in** with your LeetCode account.
 
-- You can also use the following command to sign in/out:
+- You can also use the following command to sign in/cookie in/out:
   - **LeetCode: Sign in**
+  - **LeetCode: Cookie in**
   - **LeetCode: Sign out**
 
 ---
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@
 
 - Simply click `Sign in to LeetCode` in the `LeetCode Explorer` will let you **sign in** with your LeetCode account.
 
-- You can also use the following command to sign in/cookie in/out:
+- You can also use the following command to sign in/sign in (by cookie)/out:
   - **LeetCode: Sign in**
   - **LeetCode: Sign in (by cookie)**
   - **LeetCode: Sign out**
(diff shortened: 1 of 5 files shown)


jdneo:
Is it possible to leverage the cookie extension:
 - [cookie.chrome](https://github.com/leetcode-tools/leetcode-cli-plugins/blob/master/plugins/cookie.chrome.js)
 - [cookie.firefox](https://github.com/leetcode-tools/leetcode-cli-plugins/blob/master/plugins/cookie.firefox.js)
to bypass the reCAPTCHA?

yihong0618:
Yes, but I think this way is not so simple, I will try it tomorrow.

jdneo:
@yihong0618 I just took a look about the document. The cookie plugins need more requirements and they are different for different platforms. So I think it's not an ideal solution. Just forget it.

yihong0618:
@jdneo me either. And this solution is for one-time session at that time.

yihong0618:
@jdneo 
cookie.chrome work.
But I think its not a good idea, it is no essential difference from the paste cookie method

README.md
```diff
@@ -40,8 +40,9 @@
 
 - Simply click `Sign in to LeetCode` in the `LeetCode Explorer` will let you **sign in** with your LeetCode account.
 
-- You can also use the following command to sign in/out:
+- You can also use the following command to sign in/cookie in/out:
   - **LeetCode: Sign in**
+  - **LeetCode: Cookie in**
```
jdneo: How about renaming it to `Sign in (by cookie)`

yihong0618: Of course, should I also change name in **leetcode.cookieIn command** ?

yihong0618: @jdneo 
Have changed it~

package.json
```diff
@@ -38,11 +38,17 @@
...
         "onCommand:leetcode.switchDefaultLanguage",
+        "onCommand:leetcode.signinByCookie",
         "onView:leetCodeExplorer"
     ],
     "main": "./out/src/extension",
     "contributes": {
         "commands": [
+            {
+                "command": "leetcode.signinByCookie",
+                "title": "Sign In By Cookie",
```
jdneo: nit: `By` should be lower case since it's a preposition

jdneo:
Hi @yihong0618, almost there. Just a typo.

package.json
yihong0618: Ok, I will change it right now

yihong0618: > "command": "leetcode.signinByCookie", ? 
This "by" also?

yihong0618:
Hi I fixed the typo, is that right?

jdneo:
LGTM.  Thank you @yihong0618 for your contribution! Well done!

jdneo:
Will release this fix before the end of this week

yihong0618:
Thank you~

PR #488, opened 2019-12-06
Prepare for next release

jdneo:
Add @hologerry & @yihong0618 for awareness. If there is anything missing, please feel free to let me know.


yihong0618:
@jdneo 
Can you just wait for #490 solved? I think that is not a hard problem

jdneo:
Sure. The fix should be contained in this release.

PR #496, opened 2019-12-20
add thrid party login -- GitHub and LinkedIn

yihong0618:
add thrid party login -- GitHub and LinkedIn.


jdneo:
Hi @yihong0618,

Thank you for your contribution. I have one idea about the login experience, please let me know your thoughts on this:

We still have only one entry(command) for the login, but when users trigger the command, a drop down list pops up, having the following options:
- LeetCode Account
- LeetCode Cookie
- Third-Party: GitHub
- Third-Party: LinkedIn

By doing this, the shortcut in the explorer would be applied to all kinds of login methods. And users won't need to remember different login commands. (Imagine that we might support more 3rd-party login in the future).

yihong0618:
@jdneo 
You are right I also thinked of that, can I work on it this weekend? And update the code?

jdneo:
Sure, thank you!

yihong0618:
@jdneo vscode-leetcode 2.6.20 not found ~

jdneo:
My bad! It should be available right now!

README.md
```diff
@@ -56,6 +56,9 @@ Thanks for [@yihong0618](https://github.com/yihong0618) provided a workaround wh
   - **LeetCode: Sign in (by cookie)**
   - **LeetCode: Sign out**
 
+- You can also use the following command to sign in by third party:
```
jdneo: Please only reserve the common login command as it was before.

src/leetCodeManager.ts
```diff
@@ -34,10 +34,44 @@ class LeetCodeManager extends EventEmitter {
...
+            {
+                label: "LeetCode Account",
+                description: "",
+                detail: "Use LeetCode account to login",
+                value: "LeetCode",
+            },
+            {
+                label: "LeetCode Cookie",
+                description: "",
+                detail: "Use LeetCode cookie that copy from browser to login",
```
jdneo: Use LeetCode cookie copied from browser to login

src/leetCodeManager.ts
```diff
@@ -34,10 +34,44 @@ class LeetCodeManager extends EventEmitter {
...
+                value: "Cookie",
+            },
+            {
+                label: "Third-Party: GitHub",
+                description: "",
+                detail: "Use third party GitHub account to login",
```
jdneo: Use GitHub account to login

src/leetCodeManager.ts
```diff
@@ -34,10 +34,44 @@ class LeetCodeManager extends EventEmitter {
...
+                value: "GitHub",
+            },
+            {
+                label: "Third-Party: LinkedIn",
+                description: "",
+                detail: "Use third party LinkedIn account to login",
```
jdneo: Use LinkedIn account to login

src/leetCodeManager.ts
```diff
@@ -34,10 +34,44 @@ class LeetCodeManager extends EventEmitter {
...
-    public async signIn(isByCookie: boolean = false): Promise<void> {
-        const loginArg: string = "-l";
-        const cookieArg: string = "-c";
-        const commandArg: string = isByCookie ? cookieArg : loginArg;
+    public async signIn(): Promise<void> {
+        const picks: Array<IQuickItemEx<string>> = [];
+        picks.push(
+            {
+                label: "LeetCode Account",
+                description: "",
```
jdneo: No need to set the field `description` to empty string if you do not want to render it. Just remove it from the Object.

src/leetCodeManager.ts
```diff
@@ -34,10 +34,44 @@ class LeetCodeManager extends EventEmitter {
...
+        if (!choice) {
+            return;
+        }
+        const loginMethod: string = choice.value;
+        const commandArg: string | undefined = loginCommand.get(loginMethod);
+        if (!commandArg) {
+            throw new Error(`The login method "${loginMethod}" is not supported.`);
+        }
+        const isByCookie: boolean = loginMethod === "Cookie";
         const inMessage: string = isByCookie ? "sign in by cookie" : "sign in";
```
jdneo: Is this variable `inMessage ` used anywhere?

src/shared.ts
```diff
@@ -12,6 +12,13 @@ export enum UserStatus {
     SignedOut = 2,
 }
 
+export const loginCommand: Map<string, string> = new Map([
```
jdneo: loginCommand -> loginArgsMapping

src/leetCodeManager.ts
yihong0618: Yes, it is for the difference of  cookie login or other login message.
e.g.
> promptForOpenOutputChannel(`Failed to ${inMessage}. Please open the output channel for details`

yihong0618:
@jdneo 
Thank you for your review, the naming problems had changed, and please note that it is not support for leetcode-cn.com for GitHub and LinkedIn login in leetcode-cli 2.6.20 and I raised a commit in this [pr](https://github.com/leetcode-tools/leetcode-cli/pull/35) to support it.

jdneo:
2.6.21 has been published to npm

yihong0618:
> 2.6.21 has been published to npm

Try again, two-factor seems can not support in our way in vscode-leetcode, because maybe the async to sync way in request is using read fs but in vscode terminal doesn't support. I will try to figure out another if i could. Thank you.
image
Can you just note users that we are not support two-factor in vscode-leetcode for now, if they want they can use cli to login first.
Thank you very much.

jdneo:
Thanks @yihong0618, let me also take a look first.

yihong0618:
> Thanks @yihong0618, let me also take a look first.

Thank you, that would be wonderful.
---
FYI
I am wrong that is zsh problem. 
image
If I comment this line
```javascript
                if (!pwd) {
                    childProc.kill();
                    return resolve(undefined);
                }
                childProc.stdin.write(`${pwd}\n`);
                // childProc.stdin.end();
```
vscode side can receive the "two-factor" message from cli side.
But I have problem with how to use the `vscode.window.showInputBox`

jdneo:
Sorry my laptop broke down last weekend and I have to send it to the store for repairment. I will try to resolve this PR in this week.

jdneo:
@yihong0618 

You can write something like this:

```typescript
childProc.stdout.on("data", async (data: string | Buffer) => {
    data = data.toString();
    leetCodeChannel.append(data);
    if (data.includes("two-factor")) {
        const twoFactor: string | undefined = await vscode.window.showInputBox({
            prompt: "Enter two-factor code.",
            validateInput: (s: string): string | undefined => s && s.trim() ? undefined : "The input must not be empty",
        });
        if (!twoFactor) {
            childProc.kill();
            return resolve(undefined);
        }
        childProc.stdin.write(`${twoFactor}\n`);
    } else {
        const match: RegExpMatchArray | null = data.match(/(?:.*) Successfully (login|cookie login|third party login) as (.*)/i);
        if (match && match[2]) {
            return resolve(match[2]);
        }
    }
});
```

But I found that there is some problems in the CLI side.
1. prompt-sync is declared in the `devDependencies`, which means the module will be missing in the extension side
2. Why not using `prompt` but `prompt-sync`?

yihong0618:
> @yihong0618
> 
> You can write something like this:
> 
> ```ts
> childProc.stdout.on("data", async (data: string | Buffer) => {
>     data = data.toString();
>     leetCodeChannel.append(data);
>     if (data.includes("two-factor")) {
>         const twoFactor: string | undefined = await vscode.window.showInputBox({
>             prompt: "Enter two-factor code.",
>             validateInput: (s: string): string | undefined => s && s.trim() ? undefined : "The input must not be empty",
>         });
>         if (!twoFactor) {
>             childProc.kill();
>             return resolve(undefined);
>         }
>         childProc.stdin.write(`${twoFactor}\n`);
>     } else {
>         const match: RegExpMatchArray | null = data.match(/(?:.*) Successfully (login|cookie login|third party login) as (.*)/i);
>         if (match && match[2]) {
>             return resolve(match[2]);
>         }
>     }
> });
> ```
> 
> But I found that there is some problems in the CLI side.
> 
> 1. prompt-sync is declared in the `devDependencies`, which means the module will be missing in the extension side
> 2. Why not using `prompt` but `prompt-sync`?

Because all request with session are in callback，and prompt must in the sync way. Otherwide it will fall through。
2. I will try tonight thank you
3. I also found that~ will also fix

yihong0618:
@jdneo 
Sorry, maybe I missed somthing, it didn't work on my win10 with this way.

jdneo:
@yihong0618, not quite get your point. Every async can should be able to change to sync from my understanding.

You can simply change it to:

```js
if (resp.request.uri.href !== urls.github_tf_redirect) {
  return requestLeetcodeAndSave(_request, leetcodeUrl, user, cb);
}
prompt.colors = false;
prompt.message = '';
prompt.start();
prompt.get([
  {
    name:     'code',
    required: true
  }
], function(e, result) {
  if (e) return log.fail(e);
  const authenticityTokenTwoFactor = body.match(/name="authenticity_token" value="(.*?)"/);
  if (authenticityTokenTwoFactor === null) {
    return cb('Get GitHub two-factor token failed');
  }
  const optionsTwoFactor = {
    url:     urls.github_tf_session_request,
    method:  'POST',
    headers: {
      'Content-Type': 'application/x-www-form-urlencoded',
    },
    followAllRedirects: true,
    form:               {
      'otp':                result.code,
      'authenticity_token': authenticityTokenTwoFactor[1],
      'utf8':               encodeURIComponent('✓'),
    },
  };
  _request(optionsTwoFactor, function(e, resp, body) {
    if (resp.request.uri.href === urls.github_tf_session_request) {
      return cb('Invalid two-factor code please check');
    }
    requestLeetcodeAndSave(_request, leetcodeUrl, user, cb);
  });
});
```

The `prompt-sync` is too buggy. I'm not in favor of using that module.

By using the above code, then you can update the extension side, with following two callbacks change:

```ts
childProc.stdout.on("data", async (data: string | Buffer) => {
    data = data.toString();
    leetCodeChannel.append(data);
    if (data.includes("twoFactorCode")) {
        const twoFactor: string | undefined = await vscode.window.showInputBox({
            prompt: "Enter two-factor code.",
            validateInput: (s: string): string | undefined => s && s.trim() ? undefined : "The input must not be empty",
        });
        if (!twoFactor) {
            childProc.kill();
            return resolve(undefined);
        }
        childProc.stdin.write(`${twoFactor}\n`);
        childProc.stdin.end();
    } else {
        const match: RegExpMatchArray | null = data.match(/(?:.*)Successfully .*login as (.*)/i);
        if (match && match[1]) {
            childProc.stdin.end();
            return resolve(match[1]);
        }
    }
});
```

&

```ts
childProc.on("close", (code: number) => {
    if (code !== 0) {
        reject(new Error("Failed to login.`"));
    }
});
```

yihong0618:
@jdneo 
You are right. Because I put the GitHub login without two-factor in the end of the code block, so it would fall through using prompt. And I think you are right. Really learned a lot from your project and the JS(TS), thank you. I will fix these tonight.

jdneo:
@yihong0618 No worry, I should thank you for all of the contributions you have made.

Will take a look at the PR recently.

PR #510, opened 2020-02-13
fixed use wsl feature (issue#509)

### LINKED ISSUE #509: using wsl feature not work
## 🐛 Bug Report

using wsl feature active failed
after fixe this above, it can't fetch problem list
file path conversion between windows and wsl failed during open new problem

## To Reproduce

simply active use wsl setting and reopen the extension

## Expected behavior

should work as normal

## Extension Output

some import error and "wsl node" exit code 1
then output has show the problem list while sidebar shows none of them, also a error pops up

## Your Environment

- *os*: Windows 10 Insider Build 19559
- *extension settings*: using wsl (not remote wsl, but build-in feature from extension)
- *nodejs version*: 13.8 from wsl
- *vscode version*: 1.43 insider
- *extension version*: 0.16.0

i have pointed out the bug, and fixed locally.
you converted the path to wsl path too early since the binary path is illegal

LinkeyLeo:
convert path wsl after full path constructed
resolve problem cased by wslpath conversion

leetCodeRootPathInWsl no longer used
getLeetCodeRootPath function removed since it has only one reference after the bug fixed

#509

### DIFF
--- a/src/leetCodeExecutor.ts
+++ b/src/leetCodeExecutor.ts
@@ -27,7 +27,7 @@ class LeetCodeExecutor implements Disposable {
         }, this);
     }
 
-    getLeetCodeBinaryPath() {
+    public async getLeetCodeBinaryPath() {
         return __awaiter(this, void 0, void 0, function* () {
             if (wsl.useWsl()) {
                 return `${yield wsl.toWslPath(`"${path.join(yield `"${this.leetCodeRootPath}"`, "bin", "leetcode")}"`)}`;
--- a/src/leetCodeExecutor.ts
+++ b/src/leetCodeExecutor.ts
@@ -14,13 +14,11 @@ import { toWslPath, useWsl } from "./utils/wslUtils";
 
 class LeetCodeExecutor implements Disposable {
     private leetCodeRootPath: string;
-    private leetCodeRootPathInWsl: string;
     private nodeExecutable: string;
     private configurationChangeListener: Disposable;
 
     constructor() {
         this.leetCodeRootPath = path.join(__dirname, "..", "..", "node_modules", "vsc-leetcode-cli");
-        this.leetCodeRootPathInWsl = "";
         this.nodeExecutable = this.getNodePath();
         this.configurationChangeListener = workspace.onDidChangeConfiguration((event: ConfigurationChangeEvent) => {
             if (event.affectsConfiguration("leetcode.nodePath")) {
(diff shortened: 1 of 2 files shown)


src/leetCodeExecutor.ts
```diff
@@ -29,18 +27,11 @@ class LeetCodeExecutor implements Disposable {
...
     }
 
-    public async getLeetCodeRootPath(): Promise<string> { // not wrapped by ""
+    public async getLeetCodeBinaryPath(): Promise<string> {
         if (wsl.useWsl()) {
-            if (!this.leetCodeRootPathInWsl) {
-                this.leetCodeRootPathInWsl = `${await wsl.toWslPath(this.leetCodeRootPath)}`;
-            }
-            return `${this.leetCodeRootPathInWsl}`;
+            return `${yield wsl.toWslPath(`"${path.join(yield`"${this.leetCodeRootPath}"`, "bin", "leetcode")}"`)}`;
```
jdneo: No need to have `yield` here

src/leetCodeExecutor.ts
```diff
@@ -29,18 +27,11 @@ class LeetCodeExecutor implements Disposable {
...
         }
-        return `${this.leetCodeRootPath}`;
-    }
-
-    public async getLeetCodeBinaryPath(): Promise<string> { // wrapped by ""
-        return `"${path.join(await this.getLeetCodeRootPath(), "bin", "leetcode")}"`;
+        return `"${path.join(yield`"${this.leetCodeRootPath}"`, "bin", "leetcode")}"`;
```
jdneo: No need to have `yield` here

LinkeyLeo: Orz
i'm not too much familiar with ts, maybe you should take over this fix if i keep failed the build

LinkeyLeo:
when open new problem, wslpath conversion to windows fails as the target path under wsl doesn't exists yet, I tried fixed this and append to this same PR

jdneo:
Thank you @LinkeyLeo,

Sorry I have no Windows machine right now (it's in my office, and given the current situation, It's not easy to get it).

Could you please help me to understand the change? You have mentioned:

> when open new problem, wslpath conversion to windows fails as the target path under wsl doesn't exists yet

What's the path you mentioned here?

LinkeyLeo:
> Thank you @LinkeyLeo,
> 
> Sorry I have no Windows machine right now (it's in my office, and given the current situation, It's not easy to get it).
> 
> Could you please help me to understand the change? You have mentioned:
> 
> > when open new problem, wslpath conversion to windows fails as the target path under wsl doesn't exists yet
> 
> What's the path you mentioned here?

the workflow for open problem for coding is looks like following:
1.fetchting problem
2.parsing problem title etc.
3.create new file with default codes
4.open code file and problem description in vscode

in step 3, we need to create a new source file for writing default codes, due to using node under wsl, the target file path reported by node is a wsl path, but the extension host is running under windows and need a windows path for creating and open the file, thus we need wslpath to convert the path, then the problem appeared.

wslpath will finish the path conversion only if the path is valid and the path is exists, or it rise a path doesn't exist error and won't do the path conversion.

considering other path may need to be conversion and have the same problem in the future, I fixed this problem by modified toWinPath function, making it only process the prefix of whole path and manually reconstruct the complete path

LinkeyLeo:
@jdneo 

i divide wsl paths into two parts, mounted drive root from windows which located in /mnt/ and other path located in / except /mnt/

for the after ones, wslpath need map /{WSL PATH} to \\wsl$\{distor name}\{WSL PATH}

Issue #241, opened 2019-03-27
Local debugger implementation discussion


Vigilans:
## What we've got

* LeetCode problem's `templateMeta` property from GraphQL API:
```json
    "templateMeta": {
        "name": "addTwoNumbers",
        "params": [
            {
                "name": "l1",
                "type": "ListNode",
                "dealloc": false
            },
            {
                "name": "l2",
                "type": "ListNode",
                "dealloc": false
            }
        ],
        "return": {
            "type": "ListNode",
            "dealloc": true
        }
    }
```

* Pre- and Post- Solution code template in cpp.
I've written a bunch of robust code in this [gist](https://gist.github.com/Vigilans/a40d8cebc75bd481af152307bfb8665b). To successfully build a solution, we should provide:
  1. solution code file `solution.h`.
  2. metadata of the problem `meta.h`:
```cpp
#ifndef __CLASS
#define __CLASS(name, ...)
#endif

#ifndef __METHOD
#define __METHOD(name, ...)
#endif

__CLASS(Solution);

__METHOD(inorderTraversal, vector<int>, TreeNode*);

#undef __CLASS
#undef __METHOD

#define __TREE_NODE

//#define __SYSTEM_DESIGN
```
The meta code could be generated by `templateMeta` from the extension side.

* [`vscode-cpptool-api`](https://github.com/Microsoft/vscode-cpptools-api/blob/master/api.ts) package:

This is not the `cpptool` extension, but its api package, from which we can provide custom configuration of our currently writing code.

e.g., we can force-include `pre-solution.h` in the `solution.h` to provide proper intellisense.

## What we may need

* A proper folder structure of leetcode problem files.

I suggest we keep everything about leetcode extension in `~/.leetcode`, or somewhere specified by user in settings, rather than an arbitrary workspace. All the files inside the root should be managed by extension, and user intervention should be reduced as little as possible.

* How to organize the dependency with each other language tools? Users only need a small set of languages to write leetcode solutions.

* How to trigger the debugger through a WebView, and feed the testcase data from a WebView textbox?

* ...

jdneo:
So #130 Should be one of the issues we should address first, which is to solve:
> A proper folder structure of leetcode problem files.

Vigilans:
I've created a repository for storing these bootstrapper template:
https://github.com/Vigilans/leetcode-bootstrapper

Currently it is owned by me, and when the feature is becoming mature, I will move it to @leetcode-tools.

[...]

PR #525, opened 2020-03-02
Add debug framework

wangtao0101:
Add debug framework #241 and currently support javascript and python3.


jdneo:
@wangtao0101 Thank you for the contribution. So far, I noticed some problems in this PR:

1. This is a too big PR which is almost impossible to review
2. There are some file changes which I cannot know why, like some files with number, e.g.  `src/debug/entry/javascript/problems/116.js`


Just as I have said, so far I don't want to embed the debug features into this extension, for the following reasons:
- Impossible to provide a general debug framework for all languages
- Too much uncertainties, any of the dependent language extension changes its API, we breaks.

So, my suggestion is that, you can build and publish your own extension for a specific language's debug scenario, if you need any help, I'm willing to help you.

wangtao0101:
@jdneo Thank you for your suggestion. I  will publish my own extension.

Issue #570, opened 2020-06-04
Add option to re-run test cases 


NikiforovAll:
## 🚀 Feature Proposal

Add another option to "LeetCode: Test in LeetCode" - "Run last test cases"

## Motivation
It really fits in the developer workflow. When I'm stuck I really want to fiddle the code again test cases that fail.

## Contribution
I would be glad to add PR. Do you think some static field is enough for implementation? We could also use https://code.visualstudio.com/api/references/vscode-api#Memento

Hashmap (solution Id -> test case payload)

PR #572, opened 2020-06-12
add re-run last test option

nathanesau:
re-run last option

Implement request from Issue #570


src/commands/test.ts
```diff
@@ -42,6 +42,12 @@ export async function testSolution(uri?: vscode.Uri): Promise<void> {
                 detail: "Test with the written cases in file",
                 value: ":file",
             },
+            {
+                label: "$(refresh) Re- run last...",
```
jdneo, line 46: UI: Use this icon please: `$(debug-restart)`

reference: https://microsoft.github.io/vscode-codicons/dist/codicon.html

jdneo:
Hey @nathanesau,

Thank you for your contribution, and sorry for the late reply... 

I think it could be better if we use the icon `debug-restart`

jdneo:
@nathanesau, will you still have the interest to update the PR?

nathanesau:
Sorry for the delay.

I have made the change (commit "use debug-start icon"):

re_run_last

src/commands/test.ts
```diff
@@ -42,6 +42,12 @@ export async function testSolution(uri?: vscode.Uri): Promise<void> {
                 detail: "Test with the written cases in file",
                 value: ":file",
             },
+            {
+                label: "$(debug-restart) Re-run last...",
```
jdneo, line 46: One suggestion here: We should only show the re-run option when the `lastTest` is not empty. Also, this option should put at the top of the list, to make it more convenient for the users.

PR #592, opened 2020-07-01
chore: update copyright parts of license

zry656565:


jdneo:
@zry656565 I think you also need to update all the file headers

Issue #598, opened 2020-07-16
如何将pick one得到的题目加到favorite中？


FZ-c:
如题。

yihong0618:
Find id in the editor and then star on the side bar?
![image](https://user-images.githubusercontent.com/15976103/87652025-44151980-c786-11ea-808f-c7a219766189.png)

jdneo:
@yihong0618 This is a valid workaround.

I think this issue is duplicated with #549 

Close it for duplication

FZ-c:
> Find id in the editor and then star on the side bar?
> ![image](https://user-images.githubusercontent.com/15976103/87652025-44151980-c786-11ea-808f-c7a219766189.png)

It may not be convenient, but it works. thank you~

Issue #549, opened 2020-05-09
Can you add "favorite" shortcut in .cpp, like submit , test , solution


zpeng1997:
## 🚀 Feature Proposal

It's hard to find the leetcode problem everytime in problem list and add it as favorite problem.

thx!

yihong0618:
I will try it this weekend maybe~

jdneo:
Resolved by #601

PR #601, opened 2020-07-19
add: star command in shortcuts

yihong0618:
@jdneo @zry656565 
Support star/unstar in codelen shortcuts/
image


yihong0618:
Support #598 #549

jdneo:
Thanks @yihong0618

I would suggest to use `$(star-empty)` and `$(star-full)`. You can find the style here: https://microsoft.github.io/vscode-codicons/dist/codicon.html

yihong0618:
@jdneo 
Seems a bit low then the text title(Submit, Test)?
And is it necessary to change heart icon to star icon too?
image

jdneo:
Ok, given the current rendering result, seems using text is better. But I think there is no need to put it as one of the default value

yihong0618:
> Ok, given the current rendering result, seems using text is better. But I think there is no need to put it as one of the default value

Maybe in setting that user can config it open it or not?

jdneo:
Do you mean using a setting to toggle between text and icon?

yihong0618:
> Do you mean using a setting to toggle between text and icon?

Forget it, I think for now, keep the text maybe a better way.

jdneo:
I think so too.

package.json
```diff
@@ -621,7 +621,8 @@
                         "type": "array",
                         "default": [
                             "submit",
-                            "test"
+                            "test",
+                            "star"
```
jdneo, line 625: No need to add this shortcut by default. But please add it in the items, and also documents in readme

yihong0618:
@jdneo  Fixed.

README.md
```diff
@@ -90,9 +90,10 @@ Thanks for [@yihong0618](https://github.com/yihong0618) provided a workaround wh
   <img src="https://raw.githubusercontent.com/LeetCode-OpenSource/vscode-leetcode/master/docs/imgs/shortcuts.png" alt="Editor Shortcuts" />
 </p>
 
-- The extension supports 4 editor shortcuts (aka Code Lens):
+- The extension supports 5 editor shortcuts (aka Code Lens):
   - `Submit`: Submit your answer to LeetCode.
   - `Test`: Test your answer with customized test cases.
+  - `Star/Unstar`: Star or Unstar the current problem.
```
jdneo, line 96: nit: unstar

package.json
```diff
@@ -629,6 +629,7 @@
                             "enum": [
```
jdneo, line 629: Could you by the way add `"enumDescriptions"` for this setting here? 😄

yihong0618: Fixed, and optimize star logic.

Issue #593, opened 2020-07-01
fail to install plugin


wnykuang:
## 🐛 Bug Report

Extension initialization failed. 

## To Reproduce
1. uninstall leetcode-vscode
2. reinstall leetcode-vscode
3. run the plugin, you will see it.
## Expected behavior

leetcode-vscode should be ran smoothly.

## Extension Output

v12.18.2
[WARN] Installing missing plugins, might take a while ...
- Downloading https://raw.githubusercontent.com/leetcode-tools/leetcode-cli/master/lib/plugins/github.js
[ERROR] HTTP Error: 404
[WARN] Installing missing plugins, might take a while ...
- Downloading https://raw.githubusercontent.com/leetcode-tools/leetcode-cli/master/lib/plugins/github.js
[ERROR] HTTP Error: 404
Error: Command "node "/Users/xxxxxxx/.vscode/extensions/leetcode.vscode-leetcode-0.17.0/node_modules/vsc-leetcode-cli/bin/leetcode",plugin,-i,company" failed with exit code "1".

## Your Environment

- *os*: MacOS 10.15.5
- *extension settings*: 
- *nodejs version*: node-v12.18.2
- *vscode version*: 1.46.1
- *extension version*: 0.17.0

note:
I believe the problem is caused by the 404 from 
https://raw.githubusercontent.com/leetcode-tools/leetcode-cli/master/lib/plugins/github.js

I hope it helps. :D

AndyYuan96:
I meet the similar problem yesterday, the error is 404, but I go through plugins installation, I meet 404 when list questions in leetcode-cli,  it should be related to your vpn or shadowsocks, finally I didn't solve it. Now I use the leetcode extension in clion IDE, 
just login in, nothing to wrong, you can start use.

wnykuang:
> I meet the similar problem yesterday, the error is 404, but I go through plugins installation, I meet 404 when list questions in leetcode-cli, it should be related to your vpn or shadowsocks, finally I didn't solve it. Now I use the leetcode extension in clion IDE,
> just login in, nothing to wrong, you can start use.

meh, I am not using any VPN/ssr.

AndyYuan96:
> > I meet the similar problem yesterday, the error is 404, but I go through plugins installation, I meet 404 when list questions in leetcode-cli, it should be related to your vpn or shadowsocks, finally I didn't solve it. Now I use the leetcode extension in clion IDE,
[...]

PR #602, opened 2020-07-22
fix: #593 by adding global flag to remove cache

yihong0618:


yihong0618:
@jdneo 
fix: #593 #590

src/leetCodeExecutor.ts
```diff
@@ -34,7 +36,11 @@ class LeetCodeExecutor implements Disposable {
         return `"${path.join(this.leetCodeRootPath, "bin", "leetcode")}"`;
     }
 
-    public async meetRequirements(): Promise<boolean> {
+    public async meetRequirements(context: ExtensionContext): Promise<boolean> {
+        const isUserFresh: boolean | undefined = context.globalState.get(globalStateLeetcodeIsUserFresh);
+        if (isUserFresh !== false) {
+            await this.removeOldCache();
```
jdneo, line 42: Reuse the command we have?

https://github.com/LeetCode-OpenSource/vscode-leetcode/blob/master/src/extension.ts#L51

src/leetCodeExecutor.ts
```diff
@@ -34,7 +36,11 @@ class LeetCodeExecutor implements Disposable {
         return `"${path.join(this.leetCodeRootPath, "bin", "leetcode")}"`;
     }
 
-    public async meetRequirements(): Promise<boolean> {
+    public async meetRequirements(context: ExtensionContext): Promise<boolean> {
+        const isUserFresh: boolean | undefined = context.globalState.get(globalStateLeetcodeIsUserFresh);
+        if (isUserFresh !== false) {
```
jdneo, line 41: if (!isUserFresh)

src/leetCodeExecutor.ts
yihong0618: Seems a little different.The command way use leetcode-cli way not delete the whole dir, let me check if it works fine too.

src/leetCodeExecutor.ts
yihong0618: > if (!isUserFresh)

@jdneo 
Seems not the same logic if use if (!isUserFresh)

src/leetCodeExecutor.ts
yihong0618: @jdneo 
delete cache only delete the user session but not the plugin.js cache, still have that issue. So I think we can not use it..
image

src/leetCodeExecutor.ts
```diff
@@ -61,9 +67,11 @@ class LeetCodeExecutor implements Disposable {
             try { // Check plugin
                 await this.executeCommandEx(this.nodeExecutable, [await this.getLeetCodeBinaryPath(), "plugin", "-e", plugin]);
             } catch (error) { // Download plugin and activate
+                await this.removeOldCache();
                 await this.executeCommandEx(this.nodeExecutable, [await this.getLeetCodeBinaryPath(), "plugin", "-i", plugin]);
             }
         }
+        context.globalState.update(globalStateLeetcodeIsUserFresh, false);
```
yihong0618, line 74: In the end of activate()  I set it to false, in the first it should be undefined so I use `(isUserFresh !== false)` logic to check.

jdneo: Then instead of calling it `isUserFresh`, it would be better to call it `HasInited` IMO. And set it to `true` at the end of `activate()`

The aim of that is to avoid the two falsy value `undefined` and `false` has two different meanings.

src/leetCodeExecutor.ts
jdneo: Maybe we need to change the implementation of the command `delete cache`. Considering that the current implementation does not clean entriely?

src/leetCodeExecutor.ts
yihong0618: Got it.

src/leetCodeExecutor.ts
yihong0618: Yes, current one only clean the user info(include session),  but we change the `delete cache` also have one problem beacuse user may have two account one leetcode one leetcode-cn, if we change that. both will clean, but user may only want to clean the current cache.

jdneo: I see. So let's keep the implementation. Maybe later we can provide two option in this command. Something like `Session` and `All`. Anyway it's not a high priority

src/leetCodeExecutor.ts
```diff
@@ -84,7 +85,7 @@ class LeetCodeExecutor implements Disposable {
     }
 
     public async signOut(): Promise<string> {
-        return await await this.executeCommandEx(this.nodeExecutable, [await this.getLeetCodeBinaryPath(), "user", "-L"]);
```
yihong0618, line 79: It seems that its an old typo so I changed it too, am I right?

jdneo: yes

src/shared.ts
```diff
@@ -114,3 +114,5 @@ export enum DescriptionConfiguration {
     Both = "Both",
     None = "None",
 }
+
+export const globalStateLeetcodeHasInited: string = "leetcode.isUserFresh";
```
jdneo, line 118: Please also change the value

yihong0618: Got it~!

yihong0618: Fixed.

PR #613, opened 2020-08-08
Feature: Specify test file in settings

### LINKED ISSUE #499: Specifying test files
When testing, there is an option: test with the wrietten cases in file. This requires me to select a file. How can I specify a default file? I don't want to select it every time.
Thanks.

j1george:
Added new feature to run specified test file and ability to run from a separate shortcut.


yejingxin:
Thanks for adding this feature, this is super helpful

PR #690, opened 2021-03-22
feature: Added an option to disable endpoint translation

dzz007:
After my pull request gets merged on leetcode-cli (which provides the -T option to disable endpoint translation), this pull requests will give user an option to toggle in settings whether or not they want to use endpoint's translation for the problems.


dzz007:
@jdneo Thank you for bumping leetcode-cli to 2.8.0! Do you mind publish that change to npmjs as well? So I can specify the new version in package.json. Or do you prefer to specify the commit hash in package.json?

jdneo:
@dzz007 you can first use the latest bit of the cli and link it to the vs code extension locally. Let me know if everything works fine. After that I'll release it to the registry.

dzz007:
@jdneo thank you for responding! I just tested with locally installing with master branch of leetcode-cli in vscode-leetcode. Everything works fine, I checked both the leetcode.com and leetcode-cn.com endpoints. 
The only problem I found is that changing Endpoint in vscode-leetcode's setting page actually doesn't work. But that is a separate bug which is not related to this leetcode-cli upgrade at all, and I verified that because it exists in the current version as well.

jdneo:
Just published 2.8.0.

dzz007:
@jdneo Thanks, just finished changing package.json & package-lock.json. I tested again and it works well.

jdneo:
The content of package-lock.json looks strange, why the entries have `node_modules` prefix?

dzz007:
I believe it's because I am currently using NPM v7 which introduced some changes to the lock file as specified in "lockfileVersion": 2. I can downgrade to NPM v6 and regenerate the lock file if you want.

dzz007:
Just a sidenote, IIRC v2 version of the lock-file is backward compatible with older version of NPM. So it should be able to work fine with NPMv6 for example.

jdneo:
ok, thanks for the information

dzz007:
@jdneo no problem! Do you want me to keep the lock file as-is or do you want or to use the old version?

jdneo:
If it's backward compatible, then it's fine I think.

package.json
```diff
@@ -364,6 +364,11 @@
                         ],
                         "description": "Endpoint of the user account."
                     },
+                    "leetcode.useEndpointTranslation": {
+                        "type": "boolean",
+                        "default": true,
+                        "description": "Use endpoint's translation (if available)"
```
jdneo, line 370: Please explicitly set the scope for the setting. I think it can be `application` here.

src/leetCodeExecutor.ts
```diff
@@ -88,30 +88,49 @@ class LeetCodeExecutor implements Disposable {
...
             await fse.writeFile(filePath, codeTemplate);
         }
     }
 
-    public async showSolution(input: string, language: string): Promise<string> {
-        const solution: string = await this.executeCommandWithProgressEx("Fetching top voted solution from discussions...", this.nodeExecutable, [await this.getLeetCodeBinaryPath(), "show", input, "--solution", "-l", language]);
+    public async showSolution(input: string, language: string, needTranslation: boolean): Promise<string> {
+        // solution don't support translation
+        const cmd: string[] = [await this.getLeetCodeBinaryPath(), "show", input, "--solution", "-l", language];
+        if (!needTranslation) {
```
jdneo, line 121: So the `-T` here will not take any effect?

dzz007: Currently no, it has no effect.Primarily because leetcode-cli's solution plugin is only implemented for the US end point. But I still put -T here anyways, because in the future if leetcode-cli's command supports it, it will work correctly

package.json
dzz007: Thank you for your comment! This is fixed.

dzz007:
Thank you for your review! I just added the application scope.

src/leetCodeExecutor.ts
jdneo: I'm fine to put `-T` here. It would be better to add a tsdoc for this function notifying the param `needTranslation` will not take effect.

yihong0618 (approved):

jdneo:
Coule you add a tsdoc for the `showSolution` function notifying the param needTranslation will not take effect?

dzz007:
@jdneo That's a good point! I will add it in couple hours. Do you want me to put it in leetCodeExecutor?

jdneo (approved):

dzz007:
Thanks!

jdneo:
Apologize for lazy merge. I just get the permission to merge the PRs. 😃

Issue #474, opened 2019-11-10
Can User set default language separately for different workspace?


Jarvis-ZHChen:
I got 2 workspaces for solving problems, one using typescript and the other using java. But I didn't find any setting options specific for workspace, only find setting options for user. I think it would be better to set default language for different workspace and It would be so nice of you to support some setting options (e.g. leetcode.workspaceFolder and leetcode.defaultLanguage) in current workspace.
Thanks for you reading.

Yours
Jarvis

jdneo:
Follow-up: Change the default language scope to workspace level.

2256184693:
Any solutions for this ???

Winson-Huang:
@jdneo Could you add the leetcode.workspaceFolder to workspace setting? I really hope some problems stored in one path and some problems in another path. 

If this extension can detect current folder opened in VSCode and just store problems there ignoring "leetcode.workspaceFolder", it will be very convenient.

PR #698, opened 2021-05-02
Change the configuration scope to window

adaex:
I found that the current configuration can only be used for user settings. I think this is not very friendly. For example, the workspace folder may be different on different machines. If there are multiple machines, I can only modify the workspace folder before each use.
So I changed the configuration scope to window, which can be configured in user, workspace, or remote settings.


adaex:
@jdneo  Can you think about this a little bit? I think it is necessary.

wesleyel:
[#474](https://github.com/LeetCode-OpenSource/vscode-leetcode/issues/474)

any plan to merge this?

PR #724, opened 2021-07-27
get en name from desc.url

assumecs:
One might want to view the desc in Chinese. But the file name in Chinese which auto generated will cause problems when compile/debug on Windows with g++/gdb of Cygwin or MinGW.
It can be solved like this:
``` json
"leetcode.filePath": {
  "cpp": {
    "folder": "",
    "filename": "${id}.${ext}"
  }
}
```
But maybe it's better to add an en option.
``` json
"leetcode.filePath": {
  "cpp": {
    "folder": "",
    "filename": "${id}.${name_en}.${ext}"
  }
}
```
The name_en is parsed from desc url. There might be some better source.

### DIFF
--- a/src/commands/list.ts
+++ b/src/commands/list.ts
@@ -31,6 +31,7 @@ export async function listProblems(): Promise<IProblem[]> {
                     locked: match[2].trim().length > 0,
                     state: parseProblemState(match[3]),
                     name: match[5].trim(),
+                    name_en: "",
                     difficulty: match[6].trim(),
                     passRate: match[7].trim(),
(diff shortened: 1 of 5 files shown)


jdneo:
Hi @assumecs,

Thank you for contribution. One concern is about the `name_en`. Since it might break if LeetCode changes the URL format. Do we have any other options to get the english name?

assumecs:
Hi @jdneo :
Thank you for reply.You're right, it's risky to parse the url.
I've found the property slug in leetcode-cli.It might be a better way to get the English filename #725 .

PR #725, opened 2021-07-29
Get slug from problem list for a new option of filename in English.

assumecs:
Depends on [Provide an English filename when using Endpoint Translation](https://github.com/leetcode-tools/leetcode-cli/pull/59)

This might be a better way than #724

### DIFF
--- a/src/commands/show.ts
+++ b/src/commands/show.ts
@@ -234,6 +234,8 @@ async function resolveRelativePath(relativePath: string, node: IProblem, selecte
                 return node.id;
             case "name":
                 return node.name;
+            case "slug":
+                return node.slug;
             case "camelcasename":
                 return _.camelCase(node.name);
             case "pascalcasename":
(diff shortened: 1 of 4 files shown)


yihong0618 (approved):
LGTM

src/commands/show.ts
```diff
@@ -234,6 +234,8 @@ async function resolveRelativePath(relativePath: string, node: IProblem, selecte
                 return node.id;
             case "name":
                 return node.name;
+            case "slug":
```
jdneo, line 237: I guess we need to think for a better name of this new placeholder, `slug` is not straightforward to me at least...

assumecs: You got me...
I don't know what  slug means in problem list.
Maybe name_en？

assumecs: I found this [What is a “slug” in Django](https://stackoverflow.com/questions/427102/what-is-a-slug-in-django):it is created from the title by down-casing all letters, and replacing spaces by hyphens<code>-</code>.
It's not easy to find a word to replace "slug", "short-label"?"kebab-case-title"? or just "slug"?

jdneo: Hmm, I like the idea to provide `English` name as the file name. But there is one problem: people might ask 'Ok, we have the slug styled name, so what about the camel case/pascal case/snake case'?

I think (maybe) a better solution here is having a setting to specify whether to use translated name as the file name, and then, we can provide different style option to the user(kebab, snake, pascal, ...)?

PR #727, opened 2021-08-20
Add difficulty badge to problems in explorer tree

anshkathuria:
Changes:
1. Colorize problems based on difficulty.
2. Add a badge beside the problem file in explorer tree.

Colorize-Problems


package.json
```diff
@@ -674,6 +674,35 @@
                     }
                 }
             }
+        ],
+        "colors": [
+            {
+                "id": "problems.difficulty.badge.easy",
```
jdneo, line 680: Is it possible to refer to an existing color variable? Because if we leverage an existing color variable, the color will be changed accordingly when user switch to different themes.

anshkathuria: Fixed. Now, relying on inbuilt colors.

package.json
```diff
@@ -689,7 +689,7 @@
         "@types/mocha": "^2.2.42",
         "@types/node": "^14.14.33",
         "@types/require-from-string": "^1.2.0",
-        "@types/vscode": "1.50.0",
+        "@types/vscode": "^1.56.0",
```
jdneo, line 692: Please stick to 1.56.0 here. (Remove `^`)

jdneo (commented):
Thanks for the contribution, looks pretty good!

Just one suggestion: It could be better to provide a setting here to enable user turn on/off this feature.

package.json
anshkathuria: Agreed.

package.json
```diff
@@ -671,6 +671,12 @@
                         "default": "node",
                         "scope": "application",
                         "description": "The Node.js executable path. for example, C:\\Program Files\\nodejs\\node.exe"
+                    },
+                    "leetcode.colorizeProblems": {
+                        "type": "boolean",
+                        "default": false,
```
jdneo, line 677: Let's default to `true`

And also could you please add the new setting into the readme as well? https://github.com/LeetCode-OpenSource/vscode-leetcode#settings
 
Thank you

anshkathuria: Done.

src/explorer/LeetCodeNode.ts
```diff
@@ -55,4 +55,13 @@ export class LeetCodeNode {
         };
     }
 
+    public get uri(): Uri {
+        return Uri.from({
```
jdneo, line 59: Does that mean this API is not available in 1.56.0?

anshkathuria: Yeah, its available from `1.57.0` 😞

jdneo: Haha, so let's bump engine & dependency version to 1.57.0

jdneo (approved):
LGTM. Thank you @anshkathuria

PR #728, opened 2021-08-20
Add support for sorting problems by acceptance rate

anshkathuria:
Added basic sorting strategies to sort problems based on acceptance rate. This can be extended to sort on frequency (if supported by leetcode-cli).

**Menu**
Menu

**Sorting Options**
Sorting Options

### DIFF
--- a/package.json
+++ b/package.json
@@ -690,7 +690,7 @@
                         "scope": "application",
                         "description": "Add difficulty badge and colorize problems files in explorer tree."
                     },
-                    "leetcode.problems.sort-strategy": {
+                    "leetcode.problems.sortStrategy": {
                         "type": "string",
                         "default": "None",
                         "scope": "application",
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
         "onCommand:leetcode.testSolution",
         "onCommand:leetcode.submitSolution",
         "onCommand:leetcode.switchDefaultLanguage",
+        "onCommand:leetcode.problems.sort",
         "onView:leetCodeExplorer"
     ],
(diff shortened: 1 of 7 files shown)


src/commands/plugin.ts
```diff
@@ -52,3 +53,37 @@ export function getLeetCodeEndpoint(): string {
...
+
+    const choice: IQuickItemEx<string> | undefined = await vscode.window.showQuickPick(picks);
+    if (!choice || choice.value === currentStrategy) {
+        return;
+    }
+
+    const leetCodeConfig: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("leetcode");
+    await leetCodeConfig.update("problems.sort-strategy", choice.value, true);
+    await leetCodeTreeDataProvider.refresh();
+    vscode.window.showInformationMessage(`Switched the sorting strategy to ${choice.value}`);
```
jdneo, line 83: I guess there is no need to show this information message, because the user can see the tree view refreshing immediately after switching the sorting strategy.

_// Only pop-up dialogs when necessary._

package.json
```diff
@@ -677,6 +689,17 @@
                         "default": true,
                         "scope": "application",
                         "description": "Add difficulty badge and colorize problems files in explorer tree."
+                    },
+                    "leetcode.problems.sort-strategy": {
```
jdneo, line 693: nit: sort-strategy -> sortStrategy

And please add it to the setting section in readme

jdneo (approved):
Nice job! Thank you @anshkathuria

PR #778, opened 2022-03-07
Added options to configure showing companies, tags and metadata in problem preview

dsynkd:
Also adds problem id to preview headline


dsynkd:
@jdneo @anshkathuria Could I get an approval on this please?

PR #932, opened 2024-03-13
Update package-lock

kirillyat:


kirillyat:
@jdneo it looks like extension could not get tasks and fails with
https://github.com/LeetCode-OpenSource/vscode-leetcode/issues/930


I have run nmp update to try fix this problem

leozhao21:
Now it is normal work, don't need to change `package-lock.json` file

PR #950, opened 2024-05-28
fix: premium questions are blocked

willie5588912:
I'm using [leetcode.com](https://leetcode.com/) and found that if I use a premium account and login via cookie, the "locked question" (namely, only for premium user) still jump to premium subscription web page while clicking it, not showing the question in vscode.

I guess, this may result from the failure of normal login, since it seems that there are some snippets setting the premium status, namely `GlobalState._userStatus.isPremium` during `handleUriSignIn()`.

With cookie login, the initialization stage only goes through `getLoginStatus()`.
Therefore, I add a function named `tryParseUserInfo()` to pares both username and the premium status of the user, then use `globalState.setUserStatus()` to set `isPremium`.

I've checked with both premium and non-premium accounts, both works.
Please help verify, thanks!


enzoenrico (approved):
looks good to me, great job

PR #969, opened 2024-07-29
feat: re-add cookie-based login method

tomoyachen:
Due to feedback from some users about web authorization frequently timing out and being unstable, we have decided to temporarily re-add the cookie-based login method.


leozhao21 (approved):

pandastu2212:
Id wotld

PR #972, opened 2024-08-24
[Feature] Add a small code snippet before every problem solutions

Wang-Ji20:
## About
 
  Code preamble is a short code snippet adding to the beginning of the program. A typical one is:
 
 
 ```c++
  #include <iostream>
  using namespace std;
  ```
 
  This feature is useful. Maybe inserting header files is interesting for the first several times, it is dreadful if we need to add tons of it each time we do the problems, especially for a dreadful language like C/C++
 
  ## Configuration
 
  To add a preamble for one language, you can follow this schema:
 
 ```
  "leetcode.language.preamble" = {
       "cpp" : "#include <iostream>\n",
  }
 ```

## Effect

Now you don't need to `#include <iostream>` everytime if you use cpp.


HopoZ:
where add this schema,pls?

Wang-Ji20:
> where add this schema,pls?

just in preference.. You can find a setting called "leetcode.language.preamble". Then you can edit the setting in vscode.

This is only a small utility, not very complete and did not expect it to work for everyone.

HopoZ:
> > where add this schema,pls?
> 
> just in preference.. You can find a setting called "leetcode.language.preamble". Then you can edit the setting in vscode.
> 
> This is only a small utility, not very complete and did not expect it to work for everyone.

I found a more useful method,https://github.com/LeetCode-OpenSource/vscode-leetcode/issues/469#issuecomment-612738161

Wang-Ji20:
> > > where add this schema,pls?
> > 
> > 
> > just in preference.. You can find a setting called "leetcode.language.preamble". Then you can edit the setting in vscode.
> > This is only a small utility, not very complete and did not expect it to work for everyone.
> 
> I found a more useful method,[#469 (comment)](https://github.com/LeetCode-OpenSource/vscode-leetcode/issues/469#issuecomment-612738161)

Thanks! So the cli provide this functionality, but the editor extension did not support it. no wonder I can't find relating configuration in the extension.

